PR #8, opened 2019-09-02
cpdp-521: allowed get bpr and subscription for trusts

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnector.scala
```diff
@@ -43,20 +44,56 @@ class SubscriptionConnectorImpl @Inject()(http: HttpClient, val config: Services
...
+
+object SubscriptionConnectorImpl {
+
+  sealed trait TypeOfPerson { val value: Int }
+
+  object TypeOfPerson {
+
+    final case object Individual extends TypeOfPerson {
+      val value = 1
+    }
```
co-category: where do you use this value field?

app/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnector.scala
```diff
@@ -43,20 +44,56 @@ class SubscriptionConnectorImpl @Inject()(http: HttpClient, val config: Services
...
+  }
+
+}
+
+object SubscriptionConnectorImpl {
+
+  sealed trait TypeOfPerson { val value: Int }
+
+  object TypeOfPerson {
+
```
co-category: I think a better name for this is AffinityType

test/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnectorImplSpec.scala
```diff
@@ -57,11 +58,85 @@ class SubscriptionConnectorImplSpec extends WordSpec with Matchers with MockFact
...
+        "email",
+        UkAddress("line1", None, None, None, "postcode"),
+        "sap"
+      )
+
+      val expectedRequest = Json.parse(
+        """
+          |{
+          |  "typeOfPerson" : "1",
+          |  "contactName"  : "name surname",
```
co-category: sigh - DES data modelling at it's finest.

AndyHWChung: this is subject to change - perhaps we can get a better model when we negotiate the specs with them if we ask nicely enough

app/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnector.scala
co-category: OK, this probably makes sense to keep it as TypeOfPerson if DES are expecting that field name in the JSON request.

PR #9, opened 2019-09-03
CPDP-514: add tax enrolment integration

co-category:


app/uk/gov/hmrc/cgtpropertydisposals/controllers/SubscriptionController.scala
```diff
@@ -52,7 +56,12 @@ class SubscriptionController @Inject()(service: SubscriptionService, cc: Control
                                       RequestValidationError(s"Could not parse JSON body as subscription request: $e")
                                   )
                               )
-        subscriptionResponse <- service.subscribe(subscriptionDetails).leftMap[SubscriptionError](BackendError(_))
+        subscriptionResponse <- subscriptionService
+                                 .subscribe(subscriptionDetails)
+                                 .leftMap[SubscriptionError](BackendError(_))
+        _ <- taxEnrolmentService
```
AndyHWChung: if the subscription fails we still want to see the subscription confirmation page. What will happen at the moment is that an error page will be shown since errors are mapped to a `BackendError` in the `Left`.

app/uk/gov/hmrc/cgtpropertydisposals/service/TaxEnrolmentService.scala
```diff
@@ -0,0 +1,88 @@
...
+@ImplementedBy(classOf[TaxEnrolmentServiceImpl])
+trait TaxEnrolmentService {
+  def allocateEnrolmentToGroup(cgtReference: String, subscriptionDetails: SubscriptionDetails)(
+    implicit hc: HeaderCarrier
+  ): EitherT[Future, Error, TaxEnrolmentResponse]
+}
+
+object TaxEnrolmentServiceImpl {
+
+  sealed trait TaxEnrolmentResponse
```
AndyHWChung: this would make more sense belonging in the companion object of the `TaxEnrolmentService` rather than it's implementation since the trait references this - it isn't implementation specific

app/uk/gov/hmrc/cgtpropertydisposals/controllers/SubscriptionController.scala
co-category: refactored to return 200 OK even if enrolment call fails

app/uk/gov/hmrc/cgtpropertydisposals/service/TaxEnrolmentService.scala
co-category: yes, refactored.

app/uk/gov/hmrc/cgtpropertydisposals/controllers/SubscriptionController.scala
```diff
@@ -52,24 +56,30 @@ class SubscriptionController @Inject()(service: SubscriptionService, cc: Control
...
         case RequestValidationError(msg) =>
           logger.warn(s"Error in request to subscribe: $msg")
           BadRequest
-
         case BackendError(e) =>
           logger.warn("Error while trying to subscribe", e)
           InternalServerError
+        case EnrolmentError(cgtReference) =>
+          logger.warn("Error while trying to allocate enrolment")
+          Ok(Json.toJson(cgtReference)) // We return a 200 OK even though it is an error as we will retry asynchronously
```
AndyHWChung: the only thing missing now is a unit test for this scenario

co-category: added test

co-category:
I'm wondering if we should create a separate micro-micro-service just for CGT enrolments. Might be cleaner but with the drawback of another service to maintain.

AndyHWChung:
LGTM

PR #11, opened 2019-09-05
CPDP-514: adhere to enrolment rules in service config

co-category:


app/uk/gov/hmrc/cgtpropertydisposals/service/TaxEnrolmentService.scala
```diff
@@ -60,7 +60,7 @@ class TaxEnrolmentServiceImpl @Inject()(connector: TaxEnrolmentConnector) extend
         maybePostcode match {
           case Some(postcode) =>
```
AndyHWChung: should we be sending the country code here for a non-uk address?

co-category: I'm not sure of the requirements here. I have given postcode precedence for both scenarios. Should it be country code for Non-UK addresses?

co-category: changed to always use country-code for non-uk addresses

AndyHWChung:
LGTM

PR #21, opened 2019-09-27
CPDP-515: tax enrolments retry mechanism

co-category:


test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryMongoActorSpec.scala
```diff
@@ -0,0 +1,151 @@
...
+//      mockGetAllNonFailedEnrolmentRequests()(Right(List(taxEnrolmentRequest)))
+//      taxEnrolmentRetryMongoActor ! GetAllNonFailedTaxEnrolmentRetryRequests
+//      time.advance(FiniteDuration(2, TimeUnit.SECONDS))
+//      expectMsg(
+//        RetryTaxEnrolmentRequest(
+//          taxEnrolmentRequest.copy(numOfRetries = 0, timeToNextRetryInSeconds = 0, currentElapsedTimeInSeconds = 0)
+//        )
+//      )
+//    }
+
```
co-category: I can''t get this to work. The test-actor should have received the RetryTaxEnrolmentRequest message but it doesn't actually receive anything.

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
```diff
@@ -0,0 +1,208 @@
...
+        0,
+        0,
+        0,
+        LocalDateTime.of(2019, 12, 1, 12, 10)
+      )
+
+      app ! RetryTaxEnrolmentRequest(taxEnrolmentRequest)
+      // need to advance the time
+//      time.advance(FiniteDuration(10, TimeUnit.SECONDS))
+
```
co-category: The time.advance does not seem to advance the time. Can we take a look together on this one

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
```diff
@@ -0,0 +1,133 @@
...
+import uk.gov.hmrc.cgtpropertydisposals.service.TaxEnrolmentService
+import uk.gov.hmrc.cgtpropertydisposals.util.Logging
+import uk.gov.hmrc.http.HeaderCarrier
+
+import scala.concurrent.duration.{FiniteDuration, _}
+
+class TaxEnrolmentRetryManager(
+  createMongoActor: ActorRefFactory => ActorRef,
+  createConnectorActor: ActorRefFactory => ActorRef,
+  taxEnrolmentRetryRepository: TaxEnrolmentRetryRepository,
```
co-category: I want to collapse this to one ActorRefFactor => ActorRef

co-category:
The Lock mechanism is not included in this PR. I want to get the actor implementation nailed down as I don't think it is optimal and possibly not even idiomatic. Once that is done then will have a good platform to add in the Lock mechanism.

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
```diff
@@ -0,0 +1,208 @@
...
+      val connectorProbe = TestProbe()
+
+      val app = system.actorOf(
+        Props(
+          new TaxEnrolmentRetryManager(
+            (_: ActorRefFactory) => new FakeChildMaker(self, mongoProbe.ref).newChild(),
+            (_: ActorRefFactory) => new FakeChildMaker(self, connectorProbe.ref).newChild(),
+            mockTaxEnrolmentRepository,
+            mockTaxEnrolmentService,
+            system.scheduler
```
AndyHWChung: you need to pass in time.scheduler here to make the advancing of time be picked up

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
```diff
@@ -0,0 +1,208 @@
...
+  implicit val localDateArb: Arbitrary[LocalDateTime] = Arbitrary.apply[LocalDateTime](Gen.const(LocalDateTime.now()))
+  val taxEnrolmentRequest                             = sample[TaxEnrolmentRequest]
+
+  "A Tax Enrolment Retry Manager actor" must {
+
+    "send a message to the mongo actor if tax enrolnment call was successful" in {
+      val mongoProbe     = TestProbe()
+      val connectorProbe = TestProbe()
+
+      val app = system.actorOf(
```
AndyHWChung: to avoid duplication of code in your tests as well you can factor out the common stuff in a class like this:
```scala
class TestEnvironment {
  val mongoProbe = ...
  val connectorProbe = ...
  val app = ...
}
```
and then do this in your tests:
```scala
  "blah blah blah" in new TestEnvironment {

   }
```

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
```diff
@@ -0,0 +1,208 @@
...
+        "InProgress",
+        2,
+        0,
+        249237542,
+        LocalDateTime.now()
+      )
+
+      app ! MarkTaxEnrolmentRetryRequestAsSuccess(taxEnrolmentRequest)
+
+      mongoProbe.receiveN(2)
```
AndyHWChung: you can assert the messages the mongo probe receives as well over here to make sure they're the right thing

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
co-category: I think I did try that and it didn't work as well. I will give it another go in case I missed something.

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
```diff
@@ -0,0 +1,208 @@
...
+            (_: ActorRefFactory) => new FakeChildMaker(self, mongoProbe.ref).newChild(),
+            (_: ActorRefFactory) => new FakeChildMaker(self, connectorProbe.ref).newChild(),
+            mockTaxEnrolmentRepository,
+            mockTaxEnrolmentService,
+            system.scheduler
+          )
+        )
+      )
+      connectorProbe.ref ! PoisonPill
+      expectMsg(NewChildCreated)
```
AndyHWChung: I'd include a test to see that the new child receives the messages you expect it to after it's created - you can make `NewChildCreated` include the actor ref of the new child to facilitate this

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
AndyHWChung: since there's two child you need to create this might not be possible as you'll need o know the separate ways to create then

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
```diff
@@ -0,0 +1,133 @@
...
+  taxEnrolmentService: TaxEnrolmentService,
+  scheduler: Scheduler
+) extends Actor
+    with Logging {
+  import context.dispatcher
+  implicit val hc: HeaderCarrier = HeaderCarrier()
+
+  val config: Config = ConfigFactory.load()
```
AndyHWChung: you can a `Config` from the system somewhere - something like `system.configuration`, but I can't remember exactly

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
co-category: I had a problem here. I was trying to assert over the message type but it didn't work. This probe receives two different types of messages and I am interested in the type of the second message it receives. However, the assert fails because it tries to assert on the first message it receives only. I used receiveN because I know this probe receives exactly two messages.

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryMongoActor.scala
```diff
@@ -0,0 +1,95 @@
...
+
+    case GetAllNonFailedTaxEnrolmentRetryRequests =>
+      val sender_ = sender()
+      taxEnrolmentRetryRepository.getAllNonFailedEnrolmentRequests().value.map {
+        case Left(error) => logger.warn(s"Error recovering tax enrolment retry requests: [$error]")
+        case Right(enrolmentRequests) =>
+          val rgen = new scala.util.Random
+          enrolmentRequests.foreach { enrolmentRequest =>
+            scheduler.scheduleOnce(
+              FiniteDuration(1 + rgen.nextInt(maxStaggerTime - 1) + 1, TimeUnit.SECONDS), // randomly stagger messages
```
AndyHWChung: `+1` twice is equal to `+2` - did you mean that?

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
```diff
@@ -0,0 +1,133 @@
...
+        val newConnectorActor = createConnectorActor(context)
+        context.watch(newConnectorActor)
+        context.become(initialised(mongo, newConnectorActor))
+      } else
+        logger.warn(s"Unexpected terminated message - dead actor ref: ${deadActor.toString}")
+    }
+
+    case retry: RetryTaxEnrolmentRequest =>
+      if (retry.taxEnrolmentRequest.currentElapsedTimeInSeconds >= maxAllowableElapsedTime.toSeconds) {
+        logger.info("Elapsed time has exceed maximum allowed time allowed to retry - marking request as failed")
```
AndyHWChung: including the cgt ref id in the log message might be helpful here

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryMongoActor.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  import context.dispatcher
+  implicit val hc: HeaderCarrier = HeaderCarrier()
+
+  override def receive: Receive = {
+
+    case GetAllNonFailedTaxEnrolmentRetryRequests =>
+      val sender_ = sender()
+      taxEnrolmentRetryRepository.getAllNonFailedEnrolmentRequests().value.map {
+        case Left(error) => logger.warn(s"Error recovering tax enrolment retry requests: [$error]")
```
AndyHWChung: if you import `import uk.gov.hmrc.cgtpropertydisposals.util.Logging._` you can pass the error as a second argument to `warn` and that will log the stack trace as well

app/uk/gov/hmrc/cgtpropertydisposals/models/TaxEnrolmentRequest.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import java.time.LocalDateTime
+
+import play.api.libs.json.{Json, OFormat}
+
+final case class TaxEnrolmentRequest(
+  userId: String,
+  cgtReference: String,
+  address: Address,
+  status: String                    = "InProgress",
```
AndyHWChung: if this were a sealed trait we'd know at a glance exactly what the different allowed statuses are

AndyHWChung: in `getAllNonFailedEnrolmentRequests` in `DefaultTaxEnrolmentRetryRepository`, the status that's looked for is `"Retry"` - is this different than `"InProgress"`?

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryConnectorActor.scala
```diff
@@ -0,0 +1,69 @@
...
+              } else {
+                logger.warn(
+                  s"Tax enrolment service call failed with http status: ${fail.httpResponse.status} and will not retry"
+                )
+              }
+          }
+      }
+  }
+
+  private def is5xx(status: Int): Boolean = status >= 500 && status < 600
```
AndyHWChung: `private` is a bit redundant in actor's since every method is essentially private since things outside of it can't access anything in the actor

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
```diff
@@ -0,0 +1,133 @@
...
+import uk.gov.hmrc.cgtpropertydisposals.util.Logging
+import uk.gov.hmrc.http.HeaderCarrier
+
+import scala.concurrent.duration.{FiniteDuration, _}
+
+class TaxEnrolmentRetryManager(
+  createMongoActor: ActorRefFactory => ActorRef,
+  createConnectorActor: ActorRefFactory => ActorRef,
+  taxEnrolmentRetryRepository: TaxEnrolmentRetryRepository,
+  taxEnrolmentService: TaxEnrolmentService,
```
AndyHWChung: I don't think `taxEnrolmentService` is being used in this actor

app/uk/gov/hmrc/cgtpropertydisposals/controllers/actions/AuthenticatedAction.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  private def extractHeaders(rh: RequestHeader): HeaderCarrier =
+    HeaderCarrierConverter.fromHeadersAndSessionAndRequest(rh.headers, Some(rh.session), Some(rh))
+
+  override def invokeBlock[A](request: Request[A], block: AuthenticatedRequest[A] => Future[Result]): Future[Result] = {
+    val forbidden = Results.Forbidden("Forbidden")
+    val carrier   = extractHeaders(request)
+    authorised()
+      .retrieve(v2.Retrievals.credentials) {
+        case Some(credentials) =>
```
AndyHWChung: we should be making sure it's a GG login here so that we know it's a gg cred id we're dealing with. This can be done by chking that the provider type is `GovernmentGateway`. If we get `Verify` or `PrivilegedApplication` we want to log it and turn them away with a `Forbidden`

app/uk/gov/hmrc/cgtpropertydisposals/controllers/SubscriptionController.scala
```diff
@@ -72,14 +88,34 @@ class SubscriptionController @Inject()(
...
       }
     )
   }
+
+  private def retry(taxEnrolmentRequest: TaxEnrolmentRequest): EitherT[Future, Error, Unit] =
+    EitherT[Future, Error, Unit](
+      Future {
+        Right(
+          taxEnrolmentRetryProvider.taxEnrolmentRetryManager
+            .tell(
```
AndyHWChung: if the message fails to reach the actor we're in a bit of trouble here since the enrolment will never get retried. This should probably be an ask so that you can handle an `AskTimeoutException`.

app/uk/gov/hmrc/cgtpropertydisposals/controllers/SubscriptionController.scala
```diff
@@ -72,14 +88,34 @@ class SubscriptionController @Inject()(
...
           logger.warn("Error while trying to subscribe", e)
           InternalServerError
-        case EnrolmentError(cgtReference) =>
-          logger.warn("Error while trying to allocate enrolment")
-          Ok(Json.toJson(SubscriptionResponse(cgtReference))) // We return a 200 OK even though it is an error as we will retry asynchronously
+        case FailedTaxEnrolmentCall(taxEnrolmentRequest, cgt) =>
+          logger.warn("Error while trying to allocate enrolment - sending message to retry asynchronously")
+          retry(taxEnrolmentRequest).value.map {
+            case Left(error) =>
+              logger.warn(s"Failed to submit async retry request for $taxEnrolmentRequest: error [$error]")
```
AndyHWChung: it feels like we should send an error response back to the request if we can't put the retry onto the queue - otherwise the enrolment will never get retried again. What will happen is the user will see that the subscription confirmation screen and then the next time they log in they'd be asked to subscribe again since we'd have no enrolment of them and we wouldn't be able to tell they've failed an enrolment. This'll be a confusing journey for the user.
 
If we fail the request, as far as the user is concerned they're not subscribed and will know to try again. When the user tries to subscribe again, then it'll trigger another attempt to enrol and if it fails we'll try to put it on the queue again.

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
```diff
@@ -0,0 +1,133 @@
...
+        val interval = nextScheduledTime(retry.taxEnrolmentRequest.numOfRetries)
+        val retryRequest = RetryCallToTaxEnrolmentService(
+          retry.taxEnrolmentRequest.copy(
+            numOfRetries                = retry.taxEnrolmentRequest.numOfRetries + 1,
+            timeToNextRetryInSeconds    = interval.toSeconds,
+            currentElapsedTimeInSeconds = retry.taxEnrolmentRequest.currentElapsedTimeInSeconds + retry.taxEnrolmentRequest.timeToNextRetryInSeconds
+          )
+        )
+        logger.info(s"Retrying request with following details: [${retryRequest.taxEnrolmentRequest}]")
+        scheduler.scheduleOnce(interval, connector, retryRequest)
```
AndyHWChung: if the connector actor dies before this message reaches it we're going to end up with dead letters. The simplest way to remedy this is to probably get this actor to schedule a message to itself that when received will immediately send a retry message to the connector actor.

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
co-category: I switched back to time.scheduler from the VirtualTime class and advanced the schedule but it doesn't work. I expect the connectorProbe to receive one message when the time has been advanced to represent the re-scheduling of the failed tax enrolment call but that doesn't seem to happen. I am not sure if the scheduler is being notified of the advancement in time.

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
```diff
@@ -0,0 +1,133 @@
...
+  override def preStart(): Unit = {
+    super.preStart()
+
+    val mongoActorRef     = createMongoActor(context)
+    val connectorActorRef = createConnectorActor(context)
+
+    context.watch(mongoActorRef)
+    context.watch(connectorActorRef)
+
+    mongoActorRef ! GetAllNonFailedTaxEnrolmentRetryRequests
```
AndyHWChung: if this fails to reach the mongo actor then the whole retry mechanism is broken from the start. It'd be safer here to do an ask and get acknowledgement from the mongo actor straightaway before it sends all the retry messages. If the ask times out or the getting from mongo fails then this retry manager can know to retry getting all the failed enrolment requests again

app/uk/gov/hmrc/cgtpropertydisposals/models/TaxEnrolmentRequest.scala
co-category: Good spot - Retry should not be there anymore. There are only two statuses now - InProgress and Failed. Yes, I do need to represent this as a sealed trait.

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
co-category: Neither is the TaxEnrolmentRetryRepository after the refactor - will take those out

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
```diff
@@ -0,0 +1,133 @@
...
+    context.become(initialised(mongoActorRef, connectorActorRef))
+  }
+
+  def initialised(mongo: ActorRef, connector: ActorRef): Receive = {
+    case Terminated(deadActor) => {
+      if (deadActor == mongo) {
+        logger.info("Creating new mongo actor")
+        val newMongoActor = createMongoActor(context)
```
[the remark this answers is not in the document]
AndyHWChung: also you can say something like this to shorten it:
```scala
  case Terminated(`mongo`) =>
    logger.info("Creating new mongo actor")
    context.become(initialised( context.watch(createMongoActor(context)), connector))
  case Terminated(`connector`) =>
     ...
  case Terminated(unknownActor) =>
     ...
```

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryMongoActor.scala
```diff
@@ -0,0 +1,95 @@
...
+      }
+
+    case request: DeleteTaxEnrolmentRetryRequest =>
+      taxEnrolmentRetryRepository.delete(request.taxEnrolmentRequest.userId).value.map {
+        case Left(error) => logger.warn(s"Error deleting tax enrolment retry request : [$error]")
+        case Right(count) =>
+          count match {
+            case 1 => logger.info("Successfully deleted tax enrolment retry request")
+            case _ =>
+              logger.warn(s"Could not find retry request to delete with these details: ${request.taxEnrolmentRequest}")
```
AndyHWChung: theoretically the count here could be >1 in which case the log message here would be incorrect

test/uk/gov/hmrc/cgtpropertydisposals/TestSupport.scala
```diff
@@ -0,0 +1,27 @@
...
+package uk.gov.hmrc.cgtpropertydisposals
+
+import scala.concurrent.{Await, Future}
+import scala.concurrent.duration.{Duration, FiniteDuration}
+import scala.concurrent.duration._
+
+trait TestSupport {
+  implicit val defaultTimeout: FiniteDuration = 5 seconds
+
+  def await[A](future: Future[A])(implicit timeout: Duration): A = Await.result(future, timeout)
```
AndyHWChung: we could just leverage `await` from `play.api.test.Helpers._` - it does the same thing

app/uk/gov/hmrc/cgtpropertydisposals/controllers/SubscriptionController.scala
co-category: We discussed this last Thursday. Because each request is inserted into the database before the tax enrolment request is made we don't lose information. On system recovery, the request can always be replayed. I can change it to an ask if you prefer.

co-category:
Refactoring design after discussion - closing this PR

PR #22, opened 2019-10-02
CPDP-515: simplified design and implementation of retry mechanism

co-category:


test/uk/gov/hmrc/cgtpropertydisposals/controllers/SubscriptionControllerSpec.scala
```diff
@@ -88,49 +133,109 @@ class SubscriptionControllerSpec extends ControllerSpec with ScalaCheckDrivenPro
...
+//            headerCarrier,
+//            FakeRequest().withJsonBody(Json.toJson(subscriptionDetails))
+//          )
+//
+//        val result = controller.subscribe()(request)
+//        status(result)        shouldBe OK
+//        contentAsJson(result) shouldBe Json.toJson(subscriptionResponse)
+//        //}
+//
+//      }
```
co-category: I got to fix these tests.

test/uk/gov/hmrc/cgtpropertydisposals/repositories/TaxEnrolmentRetryRepositorySpec.scala
```diff
@@ -0,0 +1,117 @@
...
+//          repository
+//            .exists(
+//              tr.userId
+//            )
+//            .value
+//        ) shouldBe (Right(Some(tr)))
+//      }
+//
+//    }
+
```
co-category: Same here.

test/uk/gov/hmrc/cgtpropertydisposals/service/TaxEnrolmentServiceImplSpec.scala
```diff
@@ -19,123 +19,88 @@ import cats.data.EitherT
...
+//        }
+//        "the http call comes back with a status of no content and the address is a non-uk address with a country code" in {
+//          mockInsertTaxEnrolmentRequestToMongoDB(taxEnrolmentRequestWithNonAddress)(Right(true))
+//          mockAllocateEnrolmentToGroup(taxEnrolmentRequestWithNonAddress)(Right(HttpResponse(204)))
+//          await(service.allocateEnrolmentToGroup(taxEnrolmentRequestWithNonAddress).value).isRight shouldBe true
+//        }
+//      }
+//    }
+//  }
 }
```
co-category: Same here. I need to work out how to mock the ask function.

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
```diff
@@ -0,0 +1,170 @@
...
+  val maxAllowableElapsedTime: FiniteDuration = config.as[FiniteDuration]("tax-enrolment-retry.max-elapsed-time")
+  val maxWaitTimeForSubmittingRecoveredEnrolmentRequests: Int =
+    config.as[Int]("tax-enrolment-retry.max-time-to-wait-after-recovery")
+
+  def is5xx(status: Int): Boolean = status >= 500 && status < 600
+
+  override def preStart(): Unit = {
+    super.preStart()
+    taxEnrolmentRetryRepository.getAllNonFailedEnrolmentRequests().value.map {
+      case Left(error) => logger.warn(s"Could not recover tax enrolment retry requests: $error")
```
AndyHWChung: if this fails, it seems like you should retry this again so that this actor can get on with retrying what it needs to

app/uk/gov/hmrc/cgtpropertydisposals/controllers/actions/AuthenticatedAction.scala
```diff
@@ -0,0 +1,67 @@
...
+    HeaderCarrierConverter.fromHeadersAndSessionAndRequest(rh.headers, Some(rh.session), Some(rh))
+
+  override def invokeBlock[A](request: Request[A], block: AuthenticatedRequest[A] => Future[Result]): Future[Result] = {
+    val forbidden = Results.Forbidden("Forbidden")
+    val carrier   = extractHeaders(request)
+    authorised()
+      .retrieve(v2.Retrievals.credentials) {
+        case Some(credentials) =>
+          val user = AuthenticatedUser(credentials.providerId)
+          block(new AuthenticatedRequest[A](user, LocalDateTime.now(), carrier, request))
```
AndyHWChung: we should be blocking users you aren't gg at this point. The `id` in `AuthenticatedUser` is thus more `ggCredId`

app/uk/gov/hmrc/cgtpropertydisposals/service/TaxEnrolmentService.scala
```diff
@@ -16,62 +16,74 @@
...
-  ): EitherT[Future, Error, TaxEnrolmentResponse] = {
-
-    val enrolmentRequest = subscriptionDetails.address match {
-      case Address.UkAddress(line1, line2, line3, line4, postcode) =>
-        EnrolmentRequest(List(KeyValuePair("Postcode", postcode)), List(KeyValuePair("CGTPDRef", cgtReference)))
-      case Address.NonUkAddress(line1, line2, line3, line4, maybePostcode, countryCode) =>
-        EnrolmentRequest(List(KeyValuePair("CountryCode", countryCode.code)), List(KeyValuePair("CGTPDRef", cgtReference)))
-    }
+  ): EitherT[Future, Error, Unit] = {
+    implicit val timeout: Timeout = Timeout(5.seconds)
```
AndyHWChung: it might be a good idea to have the timeout configurable so that you can control it in the unit test when you test what happens if there's no reply from the ask without having to wait 5 seconds

app/uk/gov/hmrc/cgtpropertydisposals/service/TaxEnrolmentService.scala
```diff
@@ -16,62 +16,74 @@
...
-            Right(TaxEnrolmentCreated)
-          case UNAUTHORIZED =>
-            Left(Error("Unauthorized"))
-          case BAD_REQUEST =>
-            Left(Error("Bad request"))
-          case other => Left(Error(s"Received unexpected http status in response from tax-enrolment service"))
+    EitherT[Future, Error, Unit](
+      taxEnrolmentRetryProvider.taxEnrolmentRetryManager
+        .ask(AttemptTaxEnrolmentAllocationToGroup(taxEnrolmentRequest))
+        .mapTo[TaxEnrolmentDatabaseInsertResponse]
```
AndyHWChung: `TaxEnrolmentDatabaseInsertResponse` as a response to `AttemptTaxEnrolmentAllocationToGroup` sounds a bit funny - I'd have expecting something like `AttemptTaxEnrolmentAllocationToGroupResponse` . Unless the request message should be `TaxEnrolmentDatabaseInsertRequest`. It just sounds like the request and response are for two different things

AndyHWChung: the database errors could be handled in the actor level and a generic error could be surfaced in the response - that way  there's less changes to make if we ever add another type of error which should result in failure

conf/application.conf
```diff
@@ -193,3 +201,13 @@ des {
     # It is possible BPR entries will list these unmatched country codes
     non-iso-country-codes = ["AN", "EU", "FC", "NT", "OR", "TP", "UN"]
 }
+
+
+
+tax-enrolment-retry {
+ min-backoff = 3.seconds
+ max-backoff = 20.seconds
+ max-elapsed-time = 60.seconds
```
AndyHWChung: I'd make these the values look a bit more realistic. If you want smaller values for tests then you can inject the config in to allow you to do that

test/uk/gov/hmrc/cgtpropertydisposals/TestSupport.scala
```diff
@@ -0,0 +1,27 @@
...
+package uk.gov.hmrc.cgtpropertydisposals
+
+import scala.concurrent.{Await, Future}
+import scala.concurrent.duration.{Duration, FiniteDuration}
+import scala.concurrent.duration._
+
+trait TestSupport {
+  implicit val defaultTimeout: FiniteDuration = 5 seconds
+
+  def await[A](future: Future[A])(implicit timeout: Duration): A = Await.result(future, timeout)
```
AndyHWChung: there's `await` from `play.api.test.Helpers._` we can use which does the same thing

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    "remove the request from db if the tax enrolment request was successful" in {
+      mockAllocateEnrolment(taxEnrolmentRequest)(Right(HttpResponse(204)))
+
+      val manager = system.actorOf(Props(new TaxEnrolmentRetryManager(connector, repository, time.scheduler)))
+
+      manager ! AttemptTaxEnrolmentAllocationToGroup(taxEnrolmentRequest)
+
+      Thread.sleep(1000) // wait for akka system to do its work
```
AndyHWChung: do you really need this `Thread.sleep`?

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  override def beforeEach: Unit =
+    mongo().drop()
+
+  override def afterAll: Unit =
+    TestKit.shutdownActorSystem(system)
+
+  val connector                  = mock[TaxEnrolmentConnector]
+  val mockRepo                   = mock[TaxEnrolmentRetryRepository]
+  val repository                 = new DefaultTaxEnrolmentRetryRepository(reactiveMongoComponent)
```
AndyHWChung: do you need an actual instance of the repo?

app/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManager.scala
```diff
@@ -0,0 +1,170 @@
...
+import scala.concurrent.duration.{FiniteDuration, _}
+
+class TaxEnrolmentRetryManager(
+  taxEnrolmentConnector: TaxEnrolmentConnector,
+  taxEnrolmentRetryRepository: TaxEnrolmentRetryRepository,
+  scheduler: Scheduler
+) extends Actor
+    with Logging {
+  import context.dispatcher
+  implicit val hc: HeaderCarrier = HeaderCarrier()
```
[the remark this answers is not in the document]
co-category: Going to investigate this today: I can see a couple of options: (1) find a way to refresh the BT if it expires (e.g. use the oauth service possibly) or (2) request a BT with no TTL and use that for the tax enrolment retry mechanism. I will talk to the platform people to see which is the best option

app/uk/gov/hmrc/cgtpropertydisposals/models/TaxEnrolmentRequest.scala
```diff
@@ -0,0 +1,55 @@
...
+  status: TaxEnrolmentRequestStatus = TaxEnrolmentInProgress,
+  timestamp: LocalDateTime          = LocalDateTime.now()
+)
+
+object TaxEnrolmentRequest {
+  sealed trait TaxEnrolmentRequestStatus extends Product with Serializable
+  case object TaxEnrolmentInProgress extends TaxEnrolmentRequestStatus
+  case object TaxEnrolmentFailed extends TaxEnrolmentRequestStatus
+
+  implicit val statusFormat: Format[TaxEnrolmentRequestStatus] = new Format[TaxEnrolmentRequestStatus] {
```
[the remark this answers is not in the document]
co-category: I switched this to use the `play-json-derived-codecs`. However one of my repo tests failed. That test selects all records with status == "TaxEnrolmentInProgress". The serialising with this codec is saving that status value to something other than that string and so the records fail to be selected. Is there a way around this? I have switched back the code to what I had before which translates each sum type to a string.

app/uk/gov/hmrc/cgtpropertydisposals/repositories/TaxEnrolmentRetryRepository.scala
```diff
@@ -0,0 +1,144 @@
...
+        .one[TaxEnrolmentRequest]
+        .map { maybeEnrolmentRequest =>
+          Right(maybeEnrolmentRequest)
+        }
+        .recover {
+          case exception => Left(Error(exception.getMessage))
+        }
+    )
+
+  override def exists(userId: String): EitherT[Future, Error, Option[TaxEnrolmentRequest]] =
```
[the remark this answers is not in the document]
co-category: I have removed exists. It's not used anywhere anymore.

test/uk/gov/hmrc/cgtpropertydisposals/actors/TaxEnrolmentRetryManagerSpec.scala
co-category: Yes as discussed yesterday. I will work on fixing this test this morning.

app/uk/gov/hmrc/cgtpropertydisposals/models/TaxEnrolmentRequest.scala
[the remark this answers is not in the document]
AndyHWChung: Could you not modify the test to look for the entries in the way that the library writes them? You could look in your local mongo store to see how it's stored. BUt I think how the library works is that if you had
```scala
sealed trait T
case class A(i: Int) extends T
case object B extends T
case class C() extends T
```
the json would look like
```
{ "A" : { "i" : 1 } }
{ "B" : { } }
{ "C" : { }  }
```

PR #26, opened 2019-10-07
used http put which odesn't throw exceptions on non2xxx responses

AndyHWChung:


co-category:
I remember the discussion we had around this last time. For non-2xx responses it does throw exceptions and this work will handle that. It's only when it blows up on connection or timeout error when this would not work.

PR #27, opened 2019-10-08
CPDP-515: add endpoint to check a users enrolment status

co-category:


app/uk/gov/hmrc/cgtpropertydisposals/service/TaxEnrolmentService.scala
```diff
@@ -72,4 +78,22 @@ class TaxEnrolmentServiceImpl @Inject()(
...
+                                .leftMap(error => Error(s"Could not check existence of enrolment in mongo: $error"))
+      taxEnrolmentRequest <- EitherT
+                              .fromOption[Future](maybeEnrolmentRequest, Error("No enrolment request found in mongo"))
+      httpResponse <- makeTaxEnrolmentCall(taxEnrolmentRequest)
+      _ <- EitherT
+            .fromEither[Future](handleTaxEnrolmentResponse(httpResponse))
+            .leftMap(error => Error(s"Could not allocate enrolment: $error"))
+      _ <- taxEnrolmentRepository
+            .delete(ggCredId)
+            .leftMap(error => Error(s"Could not delete enrolment request in mongo: $error"))
```
co-category: Is this correct? Should we return an error if the delete failed? I have asked the GGW team if the ES8 API is idempotent

AndyHWChung: should we surface an error to the user if we can't delete an entry in mongo? It won't break anything - it'll just be some wasted memory in mongo

co-category: That's my question I posted above. I have asked the GGW team about whether the ES8 API is idempotent. If it is not then this will cause problems and we will have to fix the situation where the enrolment has occurred and we didn't manage to do the delete.

app/uk/gov/hmrc/cgtpropertydisposals/service/TaxEnrolmentService.scala
```diff
@@ -34,35 +34,41 @@ trait TaxEnrolmentService {
...
 class TaxEnrolmentServiceImpl @Inject()(
   taxEnrolmentConnector: TaxEnrolmentConnector,
-  taxEnrolmentRetryRepository: TaxEnrolmentRepository
+  taxEnrolmentRepository: TaxEnrolmentRepository
 ) extends TaxEnrolmentService
     with Logging {
 
+  def makeTaxEnrolmentCall(
+    taxEnrolmentRequest: TaxEnrolmentRequest
+  )(implicit hc: HeaderCarrier): EitherT[Future, Error, HttpResponse] =
```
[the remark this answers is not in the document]
AndyHWChung: there's the presence of the `Either` in the type though which is effectively redundant - the only reason it's there is to facilitate a for-comprehension elsewhere

app/uk/gov/hmrc/cgtpropertydisposals/service/TaxEnrolmentService.scala
co-category: Removed the leftMap - we don't need want to surface this to user land.

co-category: This doesn't matter anymore. If the enrolment did occur but we are unable to delete then the call to this new endpoint will not be made and therefore duplicate tax enrolment calls will not be made.

app/uk/gov/hmrc/cgtpropertydisposals/service/TaxEnrolmentService.scala
```diff
@@ -72,4 +80,31 @@ class TaxEnrolmentServiceImpl @Inject()(
...
+    maybeTaxEnrolmentRequest: Option[TaxEnrolmentRequest]
+  )(implicit hc: HeaderCarrier): Future[Boolean] =
+    maybeTaxEnrolmentRequest match {
+      case Some(enrolmentRequest) =>
+        for {
+          result <- EitherT.liftF(makeTaxEnrolmentCall(enrolmentRequest)) // attempt enrolment
+          _      <- EitherT.fromEither(handleTaxEnrolmentResponse(result)) // evaluate enrolment result
+          _      <- taxEnrolmentRepository.delete(enrolmentRequest.ggCredId) // delete record if enrolment was successful
+        } yield ()
+        Future.successful(true) // always return true regardless of whether the tax enrolment call/delete succeeded or not
```
AndyHWChung: is there a reason you opted for this approach vs. something like:
```scala
val result = for{
  ...
} yield ()
result.leftMap.(e => log.warn(...)).merge
```
The Boolean in this method doesn't serve any useful purpose here. If you want a Boolean in the method below you could yield `maybeEnrolmentRequest.exists` which would make more sense

AndyHWChung:
LGTM

PR #29, opened 2019-10-14
CPDP-547: return CGT reference if found

co-category:


test/uk/gov/hmrc/cgtpropertydisposals/controllers/SubscriptionControllerSpec.scala
```diff
@@ -115,8 +116,9 @@ class SubscriptionControllerSpec extends ControllerSpec with ScalaCheckDrivenPro
             headerCarrier,
             FakeRequest().withJsonBody(JsString("hi"))
           )
-        mockCheckCgtEnrolmentExists(Fake.user.ggCredId)(Right(true))
-        val result = controller.checkIfUserHasASubscription()(request)
+        mockCheckCgtEnrolmentExists(Fake.user.ggCredId)(
+          Right(Some(TaxEnrolmentRequest("ggCredId", "", UkAddress("line1", None, None, None, "")))))
+        val result = controller.checkSubscriptionStatus()(request)
         status(result) shouldBe OK
```
AndyHWChung: we should check the body of the response here as well to make sure we get the right thing

co-category: added assertion

AndyHWChung:
LGTM

PR #31, opened 2019-10-15
cpdp-740: updated subscribe call to spec

AndyHWChung:


test/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnectorImplSpec.scala
```diff
@@ -107,7 +121,7 @@ class SubscriptionConnectorImplSpec extends WordSpec with Matchers with MockFact
       }
 
       "return an error when the future fails" in {
-        mockPost(s"http://host:123/subscribe", expectedHeaders, expectedRequest)(None)
+        mockPost(s"http://host:123/subscriptions/create/CGT", expectedHeaders, expectedRequest)(None)
```
co-category: extract the DES URL into a `val` to avoid duplication

co-category: also you are not doing string interpolation here so no need for the `s`

AndyHWChung: I've factored out the url now into a `val`

AndyHWChung: whoops, I did it on the wrong branch

PR #32, opened 2019-10-22
CPDP-318: display api integration

co-category:


app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionDisplayResponse.scala
```diff
@@ -0,0 +1,35 @@
...
+
+package uk.gov.hmrc.cgtpropertydisposals.models
+
+import play.api.libs.json.{Json, OFormat}
+import uk.gov.hmrc.cgtpropertydisposals.models.address.Address
+import uk.gov.hmrc.cgtpropertydisposals.models.ids.CgtReference
+import uk.gov.hmrc.cgtpropertydisposals.models.name.ContactName
+
+final case class SubscriptionDisplayResponse(
+  emailAddress: Option[Email], //TODO: DES have this as an optional field. Discuss.
```
co-category: I don't think this should be optional but in the DES schema this is not mandatory

app/uk/gov/hmrc/cgtpropertydisposals/models/des/AddressDetails.scala
```diff
@@ -0,0 +1,32 @@
...
+package uk.gov.hmrc.cgtpropertydisposals.models.des
+
+import play.api.libs.json.{Json, OFormat}
+
+final case class AddressDetails(
+  addressLine1: String,
+  addressLine2: Option[String],
+  addressLine3: Option[String],
+  addressLine4: Option[String],
+  postalCode: String, // TODO: This is a non-optional data point in the DES schema. Discuss
```
co-category: I need to double-check this

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionDisplayResponse.scala
AndyHWChung: it's officially an optional field in the spec but we will always send one over in subscription create so we can always expect one back.

AndyHWChung: I don't know whether we want to be able to test when we don't get an email address though

conf/app.routes
```diff
@@ -1,4 +1,5 @@
-POST     /business-partner-record             uk.gov.hmrc.cgtpropertydisposals.controllers.BusinessPartnerRecordController.getBusinessPartnerRecord()
-POST     /subscribe                           uk.gov.hmrc.cgtpropertydisposals.controllers.SubscriptionController.subscribe()
-POST     /register-without-id-and-subscribe   uk.gov.hmrc.cgtpropertydisposals.controllers.SubscriptionController.registerWithoutIdAndSubscribe()
-GET      /check-subscription-status           uk.gov.hmrc.cgtpropertydisposals.controllers.SubscriptionController.checkSubscriptionStatus()
\ No newline at end of file
+POST     /business-partner-record                   uk.gov.hmrc.cgtpropertydisposals.controllers.BusinessPartnerRecordController.getBusinessPartnerRecord()
+POST     /subscription                              uk.gov.hmrc.cgtpropertydisposals.controllers.SubscriptionController.subscribe()
+GET      /subscription/:cgtReference                uk.gov.hmrc.cgtpropertydisposals.controllers.SubscriptionController.getSubscription(cgtReference : String)
```
co-category: I can add a pathbinder for this so it lifts the string to a CgtReference type or shall I leave it?

app/uk/gov/hmrc/cgtpropertydisposals/models/des/AddressDetails.scala
AndyHWChung: the postcode is mandatory for uk addresses and optional for non-uk addresses. In the schema there's two definitions for the two address types. Are you looking at the uk address one?

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionDisplayResponse.scala
co-category: did we not have a discussion with them on this so they enforce the correct semantics ie we will always send this so make it non-optional

conf/app.routes
AndyHWChung: it'd be nice, but not strictly necessary. Since there's not much effort involved let's do it. Here's some code you can nick to help you https://github.com/hmrc/cgt-property-disposals/commit/06e7cb2777e0d237526c7f82b4bb292fd7fc5d49#diff-035875a9790b55d0f6b7c239a0b7b705L27

app/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnector.scala
```diff
@@ -81,6 +69,22 @@ class SubscriptionConnectorImpl @Inject()(http: HttpClient, val config: Services
...
 
+  override def getSubscription(cgtReference: CgtReference)(
+    implicit hc: HeaderCarrier
+  ): EitherT[Future, Error, HttpResponse] = {
+
+    val queryParameters = Map("regime" -> "CGT", "id" -> cgtReference.value)
+
+    EitherT[Future, Error, HttpResponse](
+      http
+        .get(baseUrl + "/subscriptions", queryParameters, headers)
```
AndyHWChung: `(baseUrl + "/subscriptions"` is essentially a static value - it could be factored outside of this method to avoid the concatenation and assignment to a temporary value every time this method is called

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionDisplayResponse.scala
```diff
@@ -0,0 +1,35 @@
...
+import uk.gov.hmrc.cgtpropertydisposals.models.ids.CgtReference
+import uk.gov.hmrc.cgtpropertydisposals.models.name.ContactName
+
+final case class SubscriptionDisplayResponse(
+  emailAddress: Option[Email],
+  address: Address,
+  contactName: ContactName,
+  cgtReference: CgtReference,
+  telephoneNumber: Option[TelephoneNumber],
+  registrationWithId: Boolean
```
AndyHWChung: could we call this `registeredWithId`? the past tense conveys the fact that the registration happened in the past

app/uk/gov/hmrc/cgtpropertydisposals/models/des/AddressDetails.scala
```diff
@@ -0,0 +1,32 @@
...
+package uk.gov.hmrc.cgtpropertydisposals.models.des
+
+import play.api.libs.json.{Json, OFormat}
+
+final case class AddressDetails(
+  addressLine1: String,
+  addressLine2: Option[String],
+  addressLine3: Option[String],
+  addressLine4: Option[String],
+  postalCode: String,
```
AndyHWChung: this should be optional - it might not be there for non-uk addresses

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionDisplayResponse.scala
co-category: Yes I also noticed that it broke the end to end manual test

app/uk/gov/hmrc/cgtpropertydisposals/models/des/AddressDetails.scala
co-category: fixed

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionDisplayResponse.scala
co-category: changed

app/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnector.scala
co-category: fixed

app/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnector.scala
```diff
@@ -81,6 +70,22 @@ class SubscriptionConnectorImpl @Inject()(http: HttpClient, val config: Services
...
 
+  override def getSubscription(cgtReference: CgtReference)(
+    implicit hc: HeaderCarrier
+  ): EitherT[Future, Error, HttpResponse] = {
+
+    val queryParameters = Map("regime" -> "CGT", "id" -> cgtReference.value)
+
+    EitherT[Future, Error, HttpResponse](
+      http
+        .get(subscriptionDisplayUrl, queryParameters, headers)
```
AndyHWChung: this needs to be updated to point to the url in the DES spec: `/subscriptions/{regime}/{idType}/{idValue}`

app/uk/gov/hmrc/cgtpropertydisposals/service/SubscriptionService.scala
```diff
@@ -53,4 +69,96 @@ class SubscriptionServiceImpl @Inject()(connector: SubscriptionConnector)(implic
...
+    desSubscriptionDisplayDetails: DesSubscriptionDisplayDetails,
+    cgtReference: CgtReference
+  ): Either[String, SubscriptionDisplayResponse] = {
+
+    val desNonIsoCountryCodes: List[CountryCode] =
+      config.underlying.get[List[CountryCode]]("des.non-iso-country-codes").value
+
+    val a = desSubscriptionDisplayDetails.subscriptionDetails.addressDetails
+
+    val addressValidation: Validation[Address] = {
```
AndyHWChung: the same code is being used in the `BusinessPartnerRecordService`. What would be nice is if we factored this out somehow, maybe something like:
```scala
object Address{
  def fromDesAddressDetails(a: DesAddressDetails)(desNonIsoCountryCodes: List[CountryCode]): ValidatedNel[String,Address] = ???
}
```

app/uk/gov/hmrc/cgtpropertydisposals/service/SubscriptionService.scala
```diff
@@ -53,4 +67,86 @@ class SubscriptionServiceImpl @Inject()(connector: SubscriptionConnector)(implic
...
+      .mapN {
+        case (address, name) =>
+          SubscriptionDisplayResponse(
+            name,
+            desSubscriptionDisplayDetails.subscriptionDetails.contactDetails.emailAddress.flatMap(e => Some(Email(e))),
+            address,
+            ContactName(desSubscriptionDisplayDetails.subscriptionDetails.contactDetails.contactName),
+            cgtReference,
+            desSubscriptionDisplayDetails.subscriptionDetails.contactDetails.phoneNumber
+              .flatMap(t => Some(TelephoneNumber(t))),
```
AndyHWChung: couldn't you just do `.map(t => Some(TelephoneNumber(t)))` instead of a `flatMap` here? Same for the email above

co-category: change it to use map instead - I like flatMap as is tells me the functor I am dealing explicitly in the code

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionDisplayResponse.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package uk.gov.hmrc.cgtpropertydisposals.models
+
+import play.api.libs.json.{Json, OFormat}
+import uk.gov.hmrc.cgtpropertydisposals.models.address.Address
+import uk.gov.hmrc.cgtpropertydisposals.models.ids.CgtReference
+import uk.gov.hmrc.cgtpropertydisposals.models.name.ContactName
+
+final case class SubscriptionDisplayResponse(
```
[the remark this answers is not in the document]
co-category: changed it

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscribedDetails.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import play.api.libs.json.{Json, OFormat}
+import uk.gov.hmrc.cgtpropertydisposals.models.address.Address
+import uk.gov.hmrc.cgtpropertydisposals.models.ids.CgtReference
+import uk.gov.hmrc.cgtpropertydisposals.models.name.{ContactName, IndividualName, TrustName}
+import uk.gov.hmrc.cgtpropertydisposals.models.EitherFormat.eitherFormat
+
+final case class SubscribedDetails(
+  name: Either[TrustName, IndividualName],
+  emailAddress: Option[Email],
```
AndyHWChung: the email address here should be non-optional. It's an error if it's not there - it's optional from DES but in our service we expect it

co-category: changed email to make it non-optional

AndyHWChung:
LGTM

PR #34, opened 2019-10-25
CPDP-816: subscription update api integration

co-category:


co-category:
@AndyHWChung do you think we should change the design of this service slightly so that we don't leak the concept of a UK or NON-UK address into it. That concept is only relevant to the front-end service. I think it will help us simplify this service and avoid messy translations between the address information coming from the frontend and the address information required by DES. Tthe frontend service should house the logic to translate between the two address encodings.

AndyHWChung:
@co-category we could do that because it'd simplify the logic somewhat in the BE. There is one thing that the BE is doing though which is DES specific which is returning a country code of `None` when DES returns a country code which is not actually an official country code. That part of the logic is DES specific and it feels like the FE shouldn't need to have knowledge about DES. At least with the way things are at the moment, any consuming service would not need to worry about working that out, or even work out if it's a UK or non-UK address.

app/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnector.scala
```diff
@@ -63,7 +67,7 @@ class SubscriptionConnectorImpl @Inject()(http: HttpClient, val config: Services
       http
         .post(subscribeUrl, Json.toJson(subscriptionRequest), headers)(
           implicitly[Writes[JsValue]],
-          hc.copy(authorization = None),
+          hc.copy(authorization = None), //FIXME: remove when DES APIs are dropped to IST0
```
AndyHWChung: we'd keep this even when DES are in their test environment - we send the `Authorization` header separately with the DES bearer token

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionUpdateRequest.scala
```diff
@@ -0,0 +1,43 @@
...
+package uk.gov.hmrc.cgtpropertydisposals.models
+
+import play.api.libs.json.{Json, OFormat}
+import uk.gov.hmrc.cgtpropertydisposals.models.EitherFormat.eitherFormat
+import uk.gov.hmrc.cgtpropertydisposals.models.SubscriptionUpdateRequest.SubscriptionUpdateDetails
+import uk.gov.hmrc.cgtpropertydisposals.models.address.Address
+import uk.gov.hmrc.cgtpropertydisposals.models.des.{ContactDetails, Individual, Trustee}
+
+final case class SubscriptionUpdateRequest(
+  regime: String,
```
AndyHWChung: shouldn't need the regime in the request to the BE - it'll always be the same when we send it to DES

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionUpdateRequest.scala
```diff
@@ -0,0 +1,43 @@
...
+  subscriptionDetails: SubscriptionUpdateDetails
+)
+
+object SubscriptionUpdateRequest {
+  implicit val format: OFormat[SubscriptionUpdateRequest] = Json.format[SubscriptionUpdateRequest]
+
+  final case class SubscriptionUpdateDetails(
```
AndyHWChung: Can we avoid mixing up the DES models with our internal models? I'd make this case class look similar to the existing `SubscriptionDetails` for instance - the FE only has the internal representations.

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionUpdateRequest.scala
co-category: removed this.

app/uk/gov/hmrc/cgtpropertydisposals/models/SubscriptionUpdateRequest.scala
co-category: yes, have made the separation

app/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnector.scala
co-category: will it need to be removed for PROD?

AndyHWChung: no, this will work in prod as well - we don't need to send the "mdtp" bearer token to des when we make the call to them

app/uk/gov/hmrc/cgtpropertydisposals/models/ContactDetails.scala
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.cgtpropertydisposals.models
+
+import play.api.libs.json.{Json, OFormat}
+
+final case class ContactDetails(
```
AndyHWChung: if we're going to start using this for the update request, we should be using it as well use it in the create subscription request and display subscription response for consistency - probably not now though since the FE will need to change as well. Unless you want to do that now?

co-category: Yeah, let's leave it for now. The next tickets are to work on the FE so could look at them then and get that done.

app/uk/gov/hmrc/cgtpropertydisposals/connectors/SubscriptionConnector.scala
AndyHWChung: can we get rid of the `FIXME`'s here?

co-category: I have removed them.

AndyHWChung:
LGTM

PR #35, opened 2019-10-28
cpdp-838: made name matching with trusts possible

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposals/models/bpr/BusinessPartnerRecordRequest.scala
```diff
@@ -40,12 +43,18 @@ object BusinessPartnerRecordRequest {
...
       case i: IndividualBusinessPartnerRecordRequest => ifIndividual(i)
     }
 
-    def id: Either[SAUTR, NINO] = fold(t => Left(t.id), _.id)
+    def foldOnId[A](
+      ifTrn: TRN => A,
+      ifSautr: SAUTR => A,
+      ifNino: NINO => A
+    ): A =
+      fold(_.id.fold(ifTrn, ifSautr), _.id.fold(ifSautr, ifNino))
```
co-category: Again, this is obfuscated code. I understand why you have done it but this can be implemented using case analysis. Granted that would be more lines of code but it would make it much easier for others to read and understand.

app/uk/gov/hmrc/cgtpropertydisposals/service/BusinessPartnerRecordService.scala
```diff
@@ -65,7 +65,7 @@ class BusinessPartnerRecordServiceImpl @Inject()(connector: BusinessPartnerRecor
     connector.getBusinessPartnerRecord(bprRequest).subflatMap { response =>
       lazy val identifiers =
         List(
-          "id"                -> bprRequest.id.fold(_.value, _.value),
+          "id"                -> bprRequest.foldOnId(_.value, _.value, _.value),
```
co-category: Since we don't document code, the code itself should be self-documenting. It is difficult to understand what this is achieving here without backtracking and find out the definition of `foldOnId`

PR #39, opened 2019-11-04
added acknowledgementReference to register without id call

AndyHWChung:


co-category:
do we need to update the stubs as well?

AndyHWChung:
good shout - the stub won't break if we don't update it, but I'll update it so that it can log the ack reference at least so we can join the calls up if we need to

PR #59, opened 2019-12-05
fix subscription and registration response events

co-category:


app/uk/gov/hmrc/cgtpropertydisposals/service/onboarding/AuditService.scala
```diff
@@ -70,38 +73,64 @@ class AuditServiceImpl @Inject()(
...
+    Try {
+      Json.parse(body)
+    } match {
+      case Failure(exception) => {
+        logger.warn(s"{Could not send subscription response audit message: $exception}")
+        sendEvent(
+          "subscriptionResponse",
+          SubscriptionResponseEvent(
+            httpStatus,
+            Json.parse("""{"body" : "failed to get subscription response"}""")
```
AndyHWChung: could we log the body that did come back? something like `"""{"body" : "could not parse body as JSON: $body"}"""` Then   at least we can see what came back. Same thing for the registration response

app/uk/gov/hmrc/cgtpropertydisposals/service/onboarding/AuditService.scala
```diff
@@ -70,38 +73,64 @@ class AuditServiceImpl @Inject()(
...
-      hc.toAuditTags("subscription-response", path)
-    )
-
-  }
+  )(implicit hc: HeaderCarrier, executionContext: ExecutionContext): Unit =
+    Try {
+      Json.parse(body)
+    } match {
+      case Failure(exception) => {
+        logger.warn(s"{Could not send subscription response audit message: $exception}")
```
AndyHWChung: doesn't look like what this log message is saying is true since we're sending an event below. We don't really care what went wrong with trying to read the body as JSON - it'll become apparent in the splunk event anyways

app/uk/gov/hmrc/cgtpropertydisposals/service/onboarding/AuditService.scala
co-category: refactored

app/uk/gov/hmrc/cgtpropertydisposals/service/onboarding/AuditService.scala
co-category: log message taken out

AndyHWChung:
LGTM

PR #68, opened 2020-01-22
cpdp-871: subscription status api

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposals/models/des/SubscriptionStatus.scala
```diff
@@ -0,0 +1,72 @@
...
+
+package uk.gov.hmrc.cgtpropertydisposals.models.des
+
+import play.api.libs.json.{JsError, JsString, JsSuccess, Reads}
+
+sealed trait SubscriptionStatus extends Product with Serializable
+
+object SubscriptionStatus {
+
+  final case object NotSubscribed extends SubscriptionStatus
```
co-category: don't need `final` here - members of objects are all final except case classes.

AndyHWChung: I've got rid of it now

PR #69, opened 2020-01-23
CPDP-1085: integrate PDF generator service

co-category:
With respect to testing, we can have a test-only controller method that writes the bytes from the PDF generator service to disk. The resultant file can then be opened using a PDF viewer. 

What is more important than that is to have a mechanism to view the PDF during development. The UX designers can do iterative design by running the PDF_GENERATOR_SERVICE locally using service-manager and then firing their HTML templates via POSTMAN to it. They will then receive the generated PDF which they can view and assess.

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposals/connectors/PdfGeneratorConnector.scala
+++ b/app/uk/gov/hmrc/cgtpropertydisposals/connectors/PdfGeneratorConnector.scala
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2020 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


app/uk/gov/hmrc/cgtpropertydisposals/connectors/PdfGeneratorConnector.scala
```diff
@@ -0,0 +1,62 @@
...
+  private val url: String          = s"$baseURL/pdf-generator-service/generate"
+
+  override def generatePDF(payload: Map[String, Seq[String]], headers: Seq[(String, String)])(
+    implicit hc: HeaderCarrier
+  ): EitherT[Future, Error, Stream[Char]] =
+    EitherT[Future, Error, Stream[Char]](
+      http
+        .POSTForm(url, payload, headers)
+        .map[Either[Error, Stream[Char]]] { response =>
+          if (response.status != Status.OK) {
```
AndyHWChung: you won't get to this point if you use `POSTForm` - that will throw exceptions if you get any 4xx or 5xx responses. To enable handling of errors inside the map you could stick a `postForm` method or something in `HttpClient` in the `http` package which does a similar thing to the other things that are already in there

co-category: fixed

AndyHWChung:
LGTM

PR #88, opened 2020-02-24
CPDP_1167 Extend test coverage for submit return

rszmacinski:


app/uk/gov/hmrc/cgtpropertydisposals/models/des/AddressDetails.scala
```diff
@@ -23,15 +23,15 @@ import cats.syntax.eq._
...
+import uk.gov.hmrc.cgtpropertydisposals.models.address.{Address, Country, Postcode}
 import uk.gov.hmrc.cgtpropertydisposals.service.onboarding.BusinessPartnerRecordServiceImpl.Validation
 
 final case class AddressDetails(
   addressLine1: String,
   addressLine2: Option[String],
   addressLine3: Option[String],
   addressLine4: Option[String],
-  postalCode: Option[String],
+  postalCode: Option[Postcode],
```
AndyHWChung: we should be able to put this back to `String` now to be safe

build.sbt
```diff
@@ -34,7 +34,7 @@ lazy val scoverageSettings =
     ScoverageKeys.coverageHighlighting := true,
     coverageEnabled.in(ThisBuild, Test, test) := true
   )
-
+resolvers += "emueller-bintray" at "http://dl.bintray.com/emueller/maven"
```
AndyHWChung: on line 60 there's a place where a resolver is already being added. You could do something like `resolvers +== Seq(...)` there so it's all in one place

test/resources/schema.json
```diff
@@ -0,0 +1,834 @@
+{
```
AndyHWChung: can we name the file something like `submit-return-des-schema-v-1-0-0` so it's more obvious what it is? If we end up using  more schemas it'll help with that. Also, it's nice to know the version of the spec the schema came from

AndyHWChung: also, when you run the service can the `countries.json`  still be found since you moved it to the test directory?

test/uk/gov/hmrc/cgtpropertydisposals/connectors/returns/SubmitReturnsConnectorSpec.scala
```diff
@@ -56,25 +63,49 @@ class SubmitReturnsConnectorSpec extends WordSpec with Matchers with MockFactory
...
-  }
+
 
   val connector = new SubmitReturnsConnectorImpl(mockHttp, new ServicesConfig(config, new RunMode(config, Mode.Test)))
 
   "SubmitReturnsConnectorImpl" when {
 
     implicit val hc: HeaderCarrier = HeaderCarrier()
-    val expectedHeaders            = Map("Authorization" -> s"Bearer $desBearerToken", "Environment" -> desEnvironment)
+    val expectedHeaders = Map("Authorization" -> s"Bearer $desBearerToken")//, "Environment" -> desEnvironment)
```
AndyHWChung: aren't we getting the `Environment` header? We should be right...?

test/uk/gov/hmrc/cgtpropertydisposals/connectors/returns/SubmitReturnsConnectorSpec.scala
```diff
@@ -56,25 +63,49 @@ class SubmitReturnsConnectorSpec extends WordSpec with Matchers with MockFactory
...
+          val submitReturnRequest: SubmitReturnRequest =
+            sample[SubmitReturnRequest].copy(completeReturn = sample[CompleteReturn]
+              .copy(propertyAddress = sample[UkAddress], triageAnswers = sample[CompleteSingleDisposalTriageAnswers].copy(numberOfProperties = One)))
+          val ppdReturnDetails = PPDReturnDetails(submitReturnRequest)
+          val desSubmitReturnRequest = DesSubmitReturnRequest(ppdReturnDetails)
+          val json = Json.toJson(desSubmitReturnRequest)
+          val validationResult = validator.validate(schemaToBeValidated, json)
+
+          withClue(s"****** Validating json ******" + json + "****** with validation result ******" + validationResult) {
+            validationResult.isSuccess shouldBe true
```
AndyHWChung: is there any way to make the test print out a helpful message if it fails, e.g. something like
```scala
if(validationResult.isSuccess) ()
else fail(s"validation failed: ${validationResult.description}")
```
not sure what methods there are on `validationResult` which might help.

app/uk/gov/hmrc/cgtpropertydisposals/models/des/AddressDetails.scala
rszmacinski: done

build.sbt
rszmacinski: sorry for this one, fixed.

test/resources/schema.json
rszmacinski: schema.json renamed to submit-return-des-schema-v-1-0-0.json

rszmacinski: yes, all tests are passing

test/uk/gov/hmrc/cgtpropertydisposals/connectors/returns/SubmitReturnsConnectorSpec.scala
rszmacinski: we get it, I enabled check

AndyHWChung:
LGTM

PR #103, opened 2020-03-02
cpdp-807-1235: financial transactions, display return

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposals/service/returns/ReturnTransformerService.scala
```diff
@@ -0,0 +1,239 @@
...
+
+@ImplementedBy(classOf[ReturnTransformerServiceImpl])
+trait ReturnTransformerService {
+
+  def toCompleteReturn(desReturn: DesReturnDetails): Either[Error, CompleteReturn]
+
+}
+
+@Singleton
+class ReturnTransformerServiceImpl @Inject() (
```
AndyHWChung: this class still needs a unit test. It's going to be quite a big job though so I was going to do it in a later PR

PR #109, opened 2020-03-06
cpdp-1341: add other disposal method

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposals/models/des/returns/DesDisposalType.scala
```diff
@@ -27,9 +27,13 @@ object DesDisposalType {
```
mikail-khan: remove this `final` modifiers - not necessary

AndyHWChung: done

PR #124, opened 2020-03-17
Cpdp 1257 bug1

modivilioglu:


conf/application.conf
```diff
@@ -135,6 +133,7 @@ metrics {
     enabled = true
 }
 
+play.i18n.langs = [ "en-GB", "en", "en-US"]
```
AndyHWChung: is this being used anywhere? If anything we should probably only support `en-GB`

AndyHWChung: could we make this "en" like the FE?

PR #128, opened 2020-03-18
cpdp-1340-1343: taxable gain moved

AndyHWChung:


test/uk/gov/hmrc/cgtpropertydisposals/service/returns/transformers/ReturnTransformerServiceImplSpec.scala
```diff
@@ -675,35 +675,6 @@ class ReturnTransformerServiceImplSpec extends WordSpec with Matchers with MockF
           result.map(_.exemptionsAndLossesDetails.annualExemptAmount) shouldBe Right(AmountInPence(1234L))
         }
 
-        "there is a net loss" in {
```
AndyHWChung: we will need to read back these values eventually, but we can do that when `CompleteReturn` handles both single disposals and multiple disposals and also calculated and non-calulated journeys

PR #129, opened 2020-03-19
Cpdp 1369 calculator results

nubz:


app/uk/gov/hmrc/cgtpropertydisposals/service/returns/transformers/ReturnTransformerService.scala
```diff
@@ -159,6 +161,18 @@ class ReturnTransformerServiceImpl @Inject() (
       AmountInPence.fromPounds(singleDisposalDetails.acquisitionFees)
     )
 
+  private def constructInitialGainAnswers(
+    singleDisposalDetails: SingleDisposalDetails
+  ): Option[AmountInPenceWithSource] = {
+    val amount: Option[AmountInPence] =
+      (singleDisposalDetails.initialLoss, singleDisposalDetails.initialGain) match {
+        case (Some(loss), _) => if (loss > 0) Some(AmountInPence.fromPounds(-loss)) else Some(AmountInPence(0L))
```
AndyHWChung: we'll always get positive amounts from DES since we can only send down positive numbers

app/uk/gov/hmrc/cgtpropertydisposals/models/returns/CalculateCgtTaxDueRequest.scala
```diff
@@ -31,7 +31,8 @@ final case class CalculateCgtTaxDueRequest(
   reliefDetails: CompleteReliefDetailsAnswers,
   exemptionAndLosses: CompleteExemptionAndLossesAnswers,
   estimatedIncome: AmountInPence,
-  personalAllowance: AmountInPence
+  personalAllowance: AmountInPence,
+  initialGainOrLoss: Option[AmountInPenceWithSource]
```
AndyHWChung: this could probably be just `AmountInPence` - when requesting the calculation, that field should always be a user supplied value

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/CgtCalculationService.scala
```diff
@@ -38,7 +38,8 @@ trait CgtCalculationService {
     reliefDetails: CompleteReliefDetailsAnswers,
     exemptionAndLosses: CompleteExemptionAndLossesAnswers,
     estimatedIncome: AmountInPence,
-    personalAllowance: AmountInPence
+    personalAllowance: AmountInPence,
+    initialGainOrLossCheck: Option[AmountInPenceWithSource]
```
AndyHWChung: why is this a `check`? I'd just call is `initialGainOrLoss`

app/uk/gov/hmrc/cgtpropertydisposals/models/des/returns/DisposalDetails.scala
```diff
@@ -89,13 +89,13 @@ object DisposalDetails {
     )
   }
 
-  private def getInitialGainOrLoss(calculatedTaxDue: CalculatedTaxDue): (BigDecimal, BigDecimal) = {
-    val value = calculatedTaxDue.initialGainOrLoss
+  private def getInitialGainOrLoss(calculatedTaxDue: CalculatedTaxDue): (Option[BigDecimal], Option[BigDecimal]) = {
```
AndyHWChung: this should now not take in a `CalulatedTaxDue`, but the `initialGainOrLoss` `AmountInPence` from a `CompleteReturn`. And then we should only populate the values in DES is that field is defined. This'll probably mean you now need to stick an optional `initialGainOrLoss`field into the `CompleteReturn`

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/transformers/ReturnTransformerService.scala
nubz: are you saying this about the `if (loss > 0)` or the actual coercion to a negative when there's Some(loss)? We can do without the `if (loss > 0)` stuff because it will return zero anyway...

AndyHWChung: yeah, I was on about the `loss > 0` part - we should be able to check the value as is and just stick a minus in front of it

app/uk/gov/hmrc/cgtpropertydisposals/models/returns/CalculateCgtTaxDueRequest.scala
nubz: done

app/uk/gov/hmrc/cgtpropertydisposals/models/des/returns/DisposalDetails.scala
nubz: changed to use the CompleteReturn to find existing vals from

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/CgtCalculationService.scala
nubz: renamed it to `initialGainOrLoss` and renamed the `val` we go on to instantiate inside to `initialGainOrLossWithSource`

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/transformers/ReturnTransformerService.scala
nubz: I removed the if greater than zero checks

app/uk/gov/hmrc/cgtpropertydisposals/models/returns/CompleteReturn.scala
```diff
@@ -32,7 +33,8 @@ final case class CompleteReturn(
   acquisitionDetails: CompleteAcquisitionDetailsAnswers,
   reliefDetails: CompleteReliefDetailsAnswers,
   exemptionsAndLossesDetails: CompleteExemptionAndLossesAnswers,
-  yearToDateLiabilityAnswers: CompleteCalculatedYearToDateLiabilityAnswers
+  yearToDateLiabilityAnswers: CompleteCalculatedYearToDateLiabilityAnswers,
+  initialGainAnswers: Option[AmountInPence]
```
AndyHWChung: `initialGainOrLoss` rather than `initialGain` would be a better name here so that it's clear this value can include a loss as well

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/CgtCalculationService.scala
```diff
@@ -38,7 +38,8 @@ trait CgtCalculationService {
     reliefDetails: CompleteReliefDetailsAnswers,
     exemptionAndLosses: CompleteExemptionAndLossesAnswers,
     estimatedIncome: AmountInPence,
-    personalAllowance: AmountInPence
+    personalAllowance: AmountInPence,
+    initialGainOrLossDetails: Option[AmountInPence]
```
AndyHWChung: I don't think it's going to break anything, but it's slightly annoying that this is called `initialGainOrLossDetails` here in the trait and then `initialGainOrLoss` later on in the class

test/uk/gov/hmrc/cgtpropertydisposals/service/returns/CgtCalculationServiceImplSpec.scala
```diff
@@ -215,8 +217,8 @@ class CgtCalculationServiceImplSpec extends WordSpec with Matchers with ScalaChe
               reliefDetails      = reliefDetails
             )
 
-            result.initialGainOrLoss.value      shouldBe 100L
-            result.gainOrLossAfterReliefs.value shouldBe 0L
+            result.initialGainOrLoss.amount.value shouldBe 100L
```
AndyHWChung: it would be better to assert the source in the `AmountInPenceWithSource` as well - then we should make sure that there is at least one test testing that we get a `Calculated` and at least one for `UserSupplied`

nubz: ok I'm writing a new test for UserSupplied vals

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/CgtCalculationService.scala
nubz: Changed to initialGainOrLoss

app/uk/gov/hmrc/cgtpropertydisposals/models/returns/CompleteReturn.scala
nubz: it's been renamed

AndyHWChung:
LGTM

PR #140, opened 2020-03-25
cpdp-1400: submit multiple disposals returns

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposals/models/ListUtils.scala
```diff
@@ -0,0 +1,32 @@
...
+    def partitionWith[B, C](f: A => Either[B, C]): (List[B], List[C]) =
+      l.foldLeft(List.empty[B] -> List.empty[C]) {
+        case (acc, curr) =>
+          f(curr).fold(
+            b => (b :: acc._1) -> acc._2,
+            c => acc._1        -> (c :: acc._2)
+          )
+      }
+  }
+
```
mikail-khan: what's wrong with the standard lib partition function? :joy:

AndyHWChung: the std library partition gives you two lists with the same type as the original list. I wanted to be able to partition into two lists with different types, so I went with this thing here.

mikail-khan: So this is just to check that valid asset types are coming from DES?

AndyHWChung: it's used somewhere when transforming des returns to our complete returns in the transformer service as well

PR #143, opened 2020-03-28
Add new test-only endpoint allowing to delete a draft return by cgtRe…

rszmacinski:
…ference.

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposals/repositories/returns/DraftReturnsRepository.scala
+++ b/app/uk/gov/hmrc/cgtpropertydisposals/repositories/returns/DraftReturnsRepository.scala
@@ -80,7 +80,7 @@ class DefaultDraftReturnsRepository @Inject() (component: ReactiveMongoComponent
 
   override def delete(cgtReference: CgtReference): EitherT[Future, Error, Unit] =
     EitherT[Future, Error, Unit](
-      remove("return.cgtReference" -> cgtReference.value)
+      remove("return.cgtReference.value" -> cgtReference.value)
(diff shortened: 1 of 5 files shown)


app/uk/gov/hmrc/cgtpropertydisposals/repositories/returns/DraftReturnsRepository.scala
```diff
@@ -76,6 +78,24 @@ class DefaultDraftReturnsRepository @Inject() (component: ReactiveMongoComponent
   override def save(draftReturn: DraftReturn, cgtReference: CgtReference): EitherT[Future, Error, Unit] =
     EitherT(set(draftReturn.id.toString, DraftReturnWithCgtReference(draftReturn, cgtReference, draftReturn.id)))
 
+  override def delete(cgtReference: CgtReference): EitherT[Future, Error, Unit] =
+    EitherT[Future, Error, Unit](
+      remove("return.cgtReference" -> cgtReference.value)
```
AndyHWChung: will this delete all draft returns for that cgt reference? You could write a test for it like this to double check: https://github.com/hmrc/cgt-property-disposals/blob/master/test/uk/gov/hmrc/cgtpropertydisposals/repositories/returns/DraftReturnsRepositorySpec.scala#L68

rszmacinski: it would not, I fixed query. I will add a test as you suggested. Thx

AndyHWChung:
LGTM

PR #148, opened 2020-03-30
CPDP-1271: pick up test DMS details from config

mikail-khan:
Raising a separate PR for the app-config-* setup

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposals/service/DmsSubmissionService.scala
+++ b/app/uk/gov/hmrc/cgtpropertydisposals/service/DmsSubmissionService.scala
@@ -56,6 +56,14 @@ class DefaultDmsSubmissionService @Inject() (
     extends DmsSubmissionService
     with Logging {
 
+  def getDmsMetaConfig[A: Configs](key: String): A =
+    configuration.underlying
+      .get[A](s"microservice.services.upscan-initiate.dms.$key")
+      .value
+
+  val queue: String        = getDmsMetaConfig[String]("classification-type")
+  val businessArea: String = getDmsMetaConfig[String]("business-area")
+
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/cgtpropertydisposals/service/DmsSubmissionService.scala
```diff
@@ -56,6 +56,14 @@ class DefaultDmsSubmissionService @Inject() (
     extends DmsSubmissionService
     with Logging {
 
+  def getDmsMetaConfig[A: Configs](key: String): A =
+    configuration.underlying
+      .get[A](s"microservice.services.upscan-initiate.dms.$key")
+      .value
+
+  val queue: String        = getDmsMetaConfig[String]("classification-type")
```
AndyHWChung: are these values actually in application.conf? I can see entries for the path `dms.queue-name` and `dms.business-area`, but not under `microserice.services.upscan-initial.dms`

AndyHWChung: should `.get[A](s"microservice.services.upscan-initiate.dms.$key")` be `.get[A](s"dms.$key")` and should this `classification-type`  be `queue-name`?

mikail-khan: good spot! Yes they are but I changed the names actually. Let me fix that

mikail-khan: fixed + fixed other config typos

AndyHWChung:
LGTM

PR #149, opened 2020-03-30
CPDP-1217: B64 encode business area

mikail-khan:


conf/application.conf
```diff
@@ -247,5 +247,5 @@ subscription-status-api.enabled = true
 dms {
     s3-file-download-timeout = 2 minutes
     queue-name = "queue-name"
-    business-area = "business-area"
+    b64-business-area = "b64-business-area"
```
AndyHWChung: do we want to base 64 encode this value to a) make it more obvious that we expect an encoded string b) not get a funny looking string at the other end? You could just use the one from your spec if you wanted `YnVzaW5lc3MtYXJlYQ==`

mikail-khan: done

AndyHWChung:
LGTM

PR #153, opened 2020-03-31
CPDP-1317

sam-miele:


app/uk/gov/hmrc/cgtpropertydisposals/models/returns/SubmitReturnResponse.scala
```diff
@@ -31,6 +32,7 @@ object SubmitReturnResponse {
   final case class ReturnCharge(
     chargeReference: String,
     amount: AmountInPence,
+    processingDate: LocalDate,
```
AndyHWChung: don't think you need this field here if it's in the `SubmitReturnResponse` already

AndyHWChung:
did the unit tests not need updating?

AndyHWChung:
LGTM

PR #154, opened 2020-03-31
cpdp-1257: fixed way of inspecting result from email service

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposals/service/returns/ReturnsService.scala
```diff
@@ -176,18 +170,10 @@ class DefaultReturnsService @Inject() (
...
-            logger.warn(s"Call to send confirmation email came back with status ${httpResponse.status}")
-            Left(Error(s"Call to send confirmation email came back with status ${httpResponse.status}"))
-          }
-        }
-      }
-      .leftMap { e =>
-        metrics.submitReturnErrorCounter.inc()
-        logger.warn("Could not send confirmation email", e)
-        e
+        if (httpResponse.status === ACCEPTED)
```
AndyHWChung: this is the actual fix - everything else is just cleanup

PR #155, opened 2020-04-01
Public beta

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposals/service/onboarding/BusinessPartnerRecordService.scala
```diff
@@ -67,9 +67,6 @@ class BusinessPartnerRecordServiceImpl @Inject() (
   val desNonIsoCountryCodes: List[CountryCode] =
     config.underlying.get[List[CountryCode]]("des.non-iso-country-codes").value
 
-  val subscriptionStatusApiEnabled: Boolean =
-    config.underlying.get[Boolean]("subscription-status-api.enabled").value
-
```
mikail-khan, line 72: remind us again, why does this need to be taken out?

AndyHWChung: this was something done on the ETMP side for the private beta release, but there wasn't time to test is on the DES side so it got moved to the public beta release. But now it's all tested, it should be good to go with the release on Monday. This is a call which tells us if someone has subscribed to cgt before with the same business partner record.

PR #166, opened 2020-04-05
CPDP-1172: draft redesign of upscan api

mikail-khan:


app/uk/gov/hmrc/cgtpropertydisposals/connectors/UpscanConnector.scala
```diff
@@ -14,64 +14,80 @@
...
+//import scala.concurrent.duration._
+//import scala.concurrent.{ExecutionContext, Future}
+//
+//@ImplementedBy(classOf[UpscanConnectorImpl])
+//trait UpscanConnector {
+//  def downloadFile(url: UpscanCallBack): Future[Either[Error, FileAttachment]]
+//}
+//
+//@Singleton
+//class UpscanConnectorImpl @Inject() (playHttpClient: PlayHttpClient, config: ServicesConfig)(
```
mikail-khan, line 58: will address this when after the new upscan api is finalised

app/uk/gov/hmrc/cgtpropertydisposals/repositories/upscan/UpscanRepository.scala
```diff
@@ -0,0 +1,152 @@
...
+        _.lastError.flatMap(_.err) match {
+          case Some(error) => Left(Error(s"could not update upscan upload: $error"))
+          case None        => Right(())
+        }
+      }
+    )
+  }
+
+  override val indexName: String = "upscan-cache-ttl"
+  override val objName: String   = ""
```
mikail-khan, line 151: why do I need to supply this objName ? It complained about it when I added the CacheRepository

app/uk/gov/hmrc/cgtpropertydisposals/service/DmsSubmissionService.scala
```diff
@@ -14,113 +14,129 @@
...
+//import uk.gov.hmrc.cgtpropertydisposals.models.returns.MandatoryEvidence
+//import uk.gov.hmrc.cgtpropertydisposals.models.returns.SupportingEvidenceAnswers.SupportingEvidence
+//import uk.gov.hmrc.cgtpropertydisposals.models.upscan.UpscanCallBack
+//import uk.gov.hmrc.cgtpropertydisposals.util.Logging
+//import uk.gov.hmrc.http.HeaderCarrier
+//
+//import scala.concurrent.{ExecutionContext, Future}
+//
+//@ImplementedBy(classOf[DefaultDmsSubmissionService])
+//trait DmsSubmissionService {
```
mikail-khan, line 63: next on the list to fix - commented out as the model is different now

app/uk/gov/hmrc/cgtpropertydisposals/models/upscan/UpscanUploadStatus.scala
```diff
@@ -0,0 +1,31 @@
...
+package uk.gov.hmrc.cgtpropertydisposals.models.upscan
+
+import cats.Eq
+import julienrf.json.derived
+import play.api.libs.json.OFormat
+
+sealed trait UpscanUploadStatus
+
+object UpscanUploadStatus {
+  case object Initiate extends UpscanUploadStatus
```
AndyHWChung, line 26: is it more accurate to have `Initiated` in the past tense here?

app/uk/gov/hmrc/cgtpropertydisposals/repositories/upscan/UpscanRepository.scala
AndyHWChung: the `CacheRepository` stores things in mongo like this:
```
{
  "_id": "abc",
   "lastUpdated": "some-date-time",
   "objName": { // whatever you tell it to write into mongo //  }
}
```
so the `objName` controls the path of what you write to mongo. So in your selector above:
```scala
val selector = Json.obj(
      "draftReturnId.value"        -> draftReturnId.value,
      "upscanUploadMeta.reference" -> upscanReference.value
    )
```
you'll have to prefix the things on the right with `s"$key.blah"`


```

Comment (mikail-khan) [reply app/uk/gov/hmrc/cgtpropertydisposals/repositories/upscan/UpscanRepository.scala:150]: what would be a good key for this?

Comment (AndyHWChung) [reply app/uk/gov/hmrc/cgtpropertydisposals/repositories/upscan/UpscanRepository.scala:150]: `upscan` maybe?

File: app/uk/gov/hmrc/cgtpropertydisposals/models/upscan/UpscanRequest.scala [hunk 5]
```diff
@@ -18,8 +18,11 @@ package uk.gov.hmrc.cgtpropertydisposals.models.upscan
 
 import play.api.libs.json.Json
 
-final case class UpscanInitiateReference(value: String) extends AnyVal
+final case class UploadRequest(
```
AndyHWChung, line 21: the file name needs updating here

mikail-khan: This model is to absorb the response coming back from upscan-initiate.

mikail-khan: The filename will be in the fields map

app/uk/gov/hmrc/cgtpropertydisposals/repositories/upscan/UpscanRepository.scala
mikail-khan: done

app/uk/gov/hmrc/cgtpropertydisposals/models/upscan/UpscanUploadStatus.scala
mikail-khan: done

app/uk/gov/hmrc/cgtpropertydisposals/models/upscan/UpscanRequest.scala
AndyHWChung: no I mean the actual name of this file - the file name is `UpscanRequest` and the case class is called `UploadRequest`

mikail-khan: lol - fixed

app/uk/gov/hmrc/cgtpropertydisposals/repositories/upscan/UpscanRepository.scala
```diff
@@ -0,0 +1,151 @@
...
+      unique   = true,
+      dropDups = true
+    )
+  )
+
+  override def insert(
+    upscanUpload: UpscanUpload
+  ): EitherT[Future, Error, Unit] =
+    EitherT[Future, Error, Unit](
+      collection.insert
```
AndyHWChung, line 85: you need to use the `set` method here from `CacheRepository` to make sure that the TTL works properly with an updated last updated time.

mikail-khan: `set` instead of `insert`?

AndyHWChung: if you look in `CacheRepository` you'll see the `set` method there - so `set` rather than `collection.insert`

AndyHWChung: actually, you probably shouldn't be using CacheRepository since that resets the last updated from the last time you touched it, but here we want to remove it 7 days from when it was created. You should be able to take out CacheRepository here and just have an index above which looks like:
```scala
Index(
    key     = Seq("uploadedOn" → IndexType.Ascending),
    name    = Some("upscan-ttl"),
    options = BSONDocument("expireAfterSeconds" -> cacheTtl.toSeconds)
  )
```

app/uk/gov/hmrc/cgtpropertydisposals/repositories/CacheRepository.scala
```diff
@@ -71,18 +86,74 @@ trait CacheRepository[A] {
...
         }
         .map { i =>
           logger.warn(s"dropping $i as ttl value is incorrect for index")
-          collection.indexesManager.drop(indexName).map(_ => ())
+          collection.indexesManager.drop(cacheTtlIndexName).map(_ => ())
         }
         .getOrElse(Future.successful(()))
     }
 
+  private def toJodaDateTime(dateTime: LocalDateTime): org.joda.time.DateTime =
```
mikail-khan, line 149: why do we need this?

AndyHWChung: the TTL depends on there being a valid date type in document somewhere - in the insert method where we have:
```scala
val lastUpdated: DateTime = overrideLastUpdatedTime.map(toJodaDateTime).getOrElse(time)

val modifier = Json.obj(
        "$set" -> Json
          .obj(
            objName       -> Json.toJson(value),
            "lastUpdated" -> lastUpdated
          )
      )
```
the thing inside `Json.obj` expects a sequence of `(String,JsValueWrapper)`. There only seems to be implicit conversions from joda time classes and not java time classes that are supplied out of the box. Hence, the conversion to joda.

PR #174, opened 2020-04-08
CPDP-1477: ensure boundedness of stream when downloading S3 files

mikail-khan:
@AndyHWChung as discussed. Need to investigate the guice dependency injection issue with tests.

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposals/connectors/UpscanConnector.scala
+++ b/app/uk/gov/hmrc/cgtpropertydisposals/connectors/UpscanConnector.scala
@@ -18,6 +18,7 @@ package uk.gov.hmrc.cgtpropertydisposals.connectors
 
 import java.util.UUID
 
+import akka.actor.ActorSystem
 import akka.stream.ActorMaterializer
 import akka.stream.scaladsl.Sink
 import com.google.inject.{ImplementedBy, Inject, Singleton}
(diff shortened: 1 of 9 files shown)


app/uk/gov/hmrc/cgtpropertydisposals/connectors/UpscanConnector.scala
```diff
@@ -55,34 +40,57 @@ trait UpscanConnector {
 }
 
 @Singleton
-class UpscanConnectorImpl @Inject() (playHttpClient: PlayHttpClient, config: ServicesConfig)(
+class UpscanConnectorImpl @Inject() (
+  playHttpClient: PlayHttpClient,
+  config: ServicesConfig,
+  implicit val mat: ActorMaterializer
```
AndyHWChung, line 46: not sure if this is going to affect anything, but what happens if you stick this in the same parameter group as the execution context?

mikail-khan: I did try that as well and I still get the same issue :upside_down_face:

AndyHWChung: did you try injecting in `system: ActorSystem`?

mikail-khan: tried that as well

PR #176, opened 2020-04-14
[CPDP-1479] - Update dependencies

modivilioglu:


project/AppDependencies.scala
```diff
@@ -5,11 +5,17 @@ object AppDependencies {
...
     "uk.gov.hmrc"                              %% "simple-reactivemongo"     % "7.26.0-play-26",
-    "uk.gov.hmrc"                              %% "bootstrap-play-26"        % "1.3.0",
+    "uk.gov.hmrc"                              %% "bootstrap-play-26"        % "1.7.0",
     "org.typelevel"                            %% "cats-core"                % "2.1.0",
     "org.julienrf"                             %% "play-json-derived-codecs" % "7.0.0",
     "com.github.kxbmap"                        %% "configs"                  % "0.4.4",
     "com.googlecode.owasp-java-html-sanitizer" % "owasp-java-html-sanitizer" % "20191001.1",
+    "uk.gov.hmrc"                              %% "auth-client"              % "3.0.0-play-26",
+    "uk.gov.hmrc"                              %% "crypto"                   % "5.6.0",
+    "uk.gov.hmrc"                              %% "http-core"                % "2.4.0",
```
AndyHWChung, line 15: do we need these new depedencies - don't they just come with bootstrap play?

mikail-khan: these are transitively pulled in

modivilioglu: Yes right, but then we will not get the latest version, as their versions are not the latest, so these override them

mikail-khan: we don't need the latest version - we only use the versions that the bootstrap library was developed against. It is one of the main reasons for using the bootstrap library. If you want to bring in the those latest versions, it would be best to update the bootstrap library dependencies and then they will be brought in that way.

modivilioglu: For example bootstrap has atm 2.33.0-play-26 . But we need 3.0.0-play so even the latest version of bootstrap won't give us what we need.

mikail-khan: when you say `need`, what or who is saying that?

AndyHWChung: I think it's the catalogue which is saying that based on the bobby plugin.

We should aim to bring in the latest of the dependencies that we already have. It is then up to bootstrap play to update the dependencies to bring us up to date. If required you can post a mesage in the PlatOps slack channel to see if they have any plans to upgrade it anytime soon.

mikail-khan: it's only a problem if the bobby rules break the build which I don't think they are.

modivilioglu: So in the ticket, https://jira.tools.tax.service.gov.uk/browse/CPDP-1479, the ticket gives the link for https://catalogue.tax.service.gov.uk/service/cgt-property-disposals which mentions the latest library versions. So that is my understanding from my chat with Andy on Thursday while he was explaining the ticket, I am more than happy just to upgrade the bootstrap version only if we do not necessarily need to upgrade the libs to their versions mentioned in the link.

AndyHWChung:
LGTM

PR #183, opened 2020-04-20
BE-test-coverage: Unit tests coverage

rambabu-posa:


test/uk/gov/hmrc/cgtpropertydisposals/connectors/returns/ReturnsConnectorSpec.scala
```diff
@@ -192,6 +193,53 @@ class ReturnsConnectorSpec extends WordSpec with Matchers with MockFactory with
 
       }
 
+      "handling request to amend a return" must {
```
AndyHWChung, line 196: this endpoint was accidentally left in - we should delete the amend return endpoint in the connector. We only used it for testing - we won't use it in it's current form when we actually come to do amendments.

rambabu-posa: Do you want me remove 'amend' from connector too?

AndyHWChung: yeah, that'd be best for now - that implementation isn't valid

AndyHWChung:
LGTM

PR #187, opened 2020-04-28
cpdp-1216: capacitor/personal rep answers in submit return

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposals/service/returns/ReturnsService.scala
```diff
@@ -94,9 +94,10 @@ class DefaultReturnsService @Inject() (
     config.underlying.get[List[CountryCode]]("des.non-iso-country-codes").value
 
   override def submitReturn(
-    returnRequest: SubmitReturnRequest
+    returnRequest: SubmitReturnRequest,
+    representeeDetails: Option[RepresenteeDetails]
```
mikail-khan, line 98: Can you just explain why this method has been extended to add `RepresenteeDetails`? Not 100% with the business logic but that info is already in SubmitReturnRequest, right?

AndyHWChung: I put it in mainly because we need to be able to get a concrete id for the representee to give to DES/ETMP. The `RepresenteeAnswers` has something of type `RepresenteeRefrenceId` in it but one of the values is `NoReferenceId`. In that instance, a call to register without id and to subscribe should be made to get a cgt reference. The intention of `RepresenteeDetails` was to pass in something which would make sure that an id for the representee was definitely available when translating from our case classes to DES data structures

app/uk/gov/hmrc/cgtpropertydisposals/models/des/returns/RepresentedPersonDetails.scala
```diff
@@ -16,24 +16,53 @@
...
   email: Option[String]
 )
 
 object RepresentedPersonDetails {
+
+  def apply(representeeDetails: RepresenteeDetails): RepresentedPersonDetails = {
+    val idTypeWithValue = representeeDetails.id match {
+      case Left(sautr)          => "UTR"  -> sautr.value
+      case Right(Left(nino))    => "NINO" -> nino.value
+      case Right(Right(cgtRef)) => "ZCGT" -> cgtRef.value
```
mikail-khan, line 46: are you missing `TRN` here ?

AndyHWChung: we don't allow trusts to be represented by capacitors or personal representatives, so that wasn't included in `RepresenteeRefrenceId`. TRN in the DES spec because it was drawn up to future proof against things we didn't know at the time

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/ReturnsService.scala
```diff
@@ -61,7 +61,7 @@ import scala.util.Try
 @ImplementedBy(classOf[DefaultReturnsService])
 trait ReturnsService {
 
-  def submitReturn(returnRequest: SubmitReturnRequest)(
+  def submitReturn(returnRequest: SubmitReturnRequest, representeeDetails: Option[RepresenteeDetails])(
     implicit hc: HeaderCarrier,
```
mikail-khan, line 65: I'm just thinking is there anyway to avoid extending this method with an additional parameter - I would like to keep this method strictly a one parameter method that accepts a `returnRequest`. Extending it opens the doors to extending it ad-hoc in the future. Is there a way to extract the id and then update SubmitReturnRequest?

AndyHWChung: I thought about updating copying the representee answers with the new cgt ref, but then that doesn't avoid the scenario you have to handle where you have to handle a `NoReferenceId` when trying to translate. We know we don't have to handle that case at that point because we should always pass in a valid cgt reference in that scenario

mikail-khan: hmm..something doesn't seem right. To me, when we reach `submitReturn` all information should have been captured already.

mikail-khan:
will re-visit the PR comments later

PR #189, opened 2020-05-03
cpdp-1208: handle cgt account not found

AndyHWChung:


AndyHWChung:
to be merged in at the same time as https://github.com/hmrc/cgt-property-disposals-frontend/pull/482

PR #200, opened 2020-06-02
CPDP-1835 - Fix Compiler Warnings

modivilioglu:


build.sbt
```diff
@@ -62,9 +62,9 @@ lazy val microservice = Project(appName, file("."))
     addCompilerPlugin(scalafixSemanticdb),
     scalacOptions ++= List(
       "-Yrangepos",
-      "-Ywarn-unused:imports",
-      "-Ypartial-unification"
-    )
+      "-language:postfixOps"
```
AndyHWChung, line 65: didn't we want to enable a bunch of other compiler flags as well. These were the ones you mentioned which I copied in the ticket:
```
"-Ywarn-dead-code", // Warn when dead code is identified.
 "-Ywarn-extra-implicit", // Warn when more than one implicit parameter section is defined.
 "-Ywarn-inaccessible", // Warn about inaccessible types in method signatures.
 "-Ywarn-infer-any", // Warn when a type argument is inferred to be `Any`.
 "-Ywarn-nullary-override", // Warn when non-nullary `def f()' overrides nullary `def f'.
 "-Ywarn-nullary-unit", // Warn when nullary methods return Unit.
 "-Ywarn-numeric-widen", // Warn when numerics are widened.
 "-Ywarn-unused:implicits", // Warn if an implicit parameter is unused.
 "-Ywarn-unused:imports", // Warn if an import selector is not referenced.
 "-Ywarn-unused:locals", // Warn if a local definition is unused.
 "-Ywarn-unused:params", // Warn if a value parameter is unused.
 "-Ywarn-unused:patvars", // Warn if a variable bound in a pattern is unused.
 "-Ywarn-unused:privates", // Warn if a private member is unused.
 "-Ywarn-value-discard" // Warn when non-Unit expression results are unused.
```

mikail-khan: the plugin covers it - that's why I introduced it so we don't need to manage all these flags.

AndyHWChung:
LGTM

PR #226, opened 2020-07-29
CPDP-2208: augment returns with clearing reason

mikail-khan:


app/uk/gov/hmrc/cgtpropertydisposals/models/finance/ClearingReason.scala
```diff
@@ -0,0 +1,51 @@
...
+  case object IncomingPayment extends ClearingReason
+  case object OutgoingPayment extends ClearingReason
+  case object WriteOff extends ClearingReason
+  case object Reversal extends ClearingReason
+  case object MassWriteOff extends ClearingReason
+  case object AutomaticClearing extends ClearingReason
+  case object SomeOtherClearingReason extends ClearingReason
+  case object InvalidClearingReason extends ClearingReason
+
+  def fromString(clearingReason: String): ClearingReason =
```
AndyHWChung, line 37: it'd make more sense to have this as:
```scala
def fromString(clearingReason: String): Option[ClearingReason]
```
instead of having `InvalidClearingReason`. That's an error case and shouldn't be passed around to the FE

mikail-khan: No, because I need to assert against this case in the twirl template to then default to the payment details message. If I error out then I can't asset in the twirl template. See the FE twirl template changes. Or did I not interpret the ticket correctly?

AndyHWChung: the clearing reason being blank was meant to be the case when the clearing reason wasn't given from DES at all, which in the transformer class would be the case:
```scala
 case DesFinancialTransactionItem(Some(paymentAmount), Some(paymentMethod), Some(clearingDate), None) =>
```
Thinking about it now, that means that you could just have 
```scala
case _ => SomeOtherClearingReason
```
at the end of this pattern match and keep it non-optional. Getting an empty string for the field would be weird to get

mikail-khan: Blank when dealing with strings is not the same as missing as you know. Need to tighten up these story specs. I'll make those changes.

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/transformers/ReturnSummaryListTransformerService.scala
```diff
@@ -148,18 +149,90 @@ class ReturnSummaryListTransformerServiceImpl extends ReturnSummaryListTransform
...
+              )
+            )
+
+        case DesFinancialTransactionItem(
+              Some(_),
+              None,
+              Some(_),
+              None
+            ) =>
+          Left(s"Both payment method and clearing reason are missing.")
```
mikail-khan, line 225: @AndyHWChung do you want me to remove this and handle this case in the FE?

mikail-khan: I just read your other review comment on the FE, so I will leave this in here

AndyHWChung:
LGTM

PR #227, opened 2020-07-31
CPDP-2220: add repayment field to draft return

AndyHWChung:


test/uk/gov/hmrc/cgtpropertydisposals/controllers/returns/SubmitReturnsControllerSpec.scala
```diff
@@ -237,16 +237,16 @@ class SubmitReturnsControllerSpec extends ControllerSpec {
...
         "there is no reference id for a representee" in {
 
-            val requestBody = sample[SubmitReturnRequest].copy(
-              checkYourAnswerPageHtml = B64Html(""),
-              completeReturn = sample[CompleteSingleDisposalReturn].copy(
-                representeeAnswers = Some(
-                  sample[CompleteRepresenteeAnswers].copy(id = NoReferenceId)
-                )
+          val requestBody = sample[SubmitReturnRequest].copy(
+            checkYourAnswerPageHtml = B64Html(""),
```
mikail-khan, line 241: can we get a generator for `B64Html` or will a random string break the tests?

AndyHWChung: it'll break it I think cos it'll try to base64 decode it in the code somewhere

mikail-khan: I think it shouldn't because what's there at the moment is not valid HTML anyway and we're not doing integration tests here so it should be fine? one way to find out....

AndyHWChung: it'll probably break around here:
https://github.com/hmrc/cgt-property-disposals/blob/master/app/uk/gov/hmrc/cgtpropertydisposals/controllers/returns/SubmitReturnsController.scala#L89

AndyHWChung: we decode it in the controller to be able to sanitise the html in it

mikail-khan: pushed up some code to allow B64HTML gen

AndyHWChung: thanks, looks good

PR #229, opened 2020-08-06
CPDP-2466: add isFurtherReturn flag

mikail-khan:


app/uk/gov/hmrc/cgtpropertydisposals/models/des/returns/DesReturnDetails.scala
```diff
@@ -61,8 +63,14 @@ object DesReturnDetails {
...
-  private def getSource(submitReturnRequest: SubmitReturnRequest): String =
-    submitReturnRequest.agentReferenceNumber.fold("self digital")(_ => "agent digital")
+  private def getSource(submitReturnRequest: SubmitReturnRequest): String = {
+    val source =
+      if (submitReturnRequest.isFurtherReturn)
+        s"${"self digital " + Instant.now.getEpochSecond.toString}"
+      else
+        "self digital"
+    submitReturnRequest.agentReferenceNumber.fold(source)(_ => "agent digital")
+  }
```
mikail-khan, line 73: does "agent digital" need to have this timestamp added?

AndyHWChung: yeah, we'll need it if an agent did the return as well

app/uk/gov/hmrc/cgtpropertydisposals/models/des/returns/DesReturnDetails.scala
```diff
@@ -61,8 +63,11 @@ object DesReturnDetails {
     )
   }
 
-  private def getSource(submitReturnRequest: SubmitReturnRequest): String =
-    submitReturnRequest.agentReferenceNumber.fold("self digital")(_ => "agent digital")
+  private def getSource(submitReturnRequest: SubmitReturnRequest): String = {
+    val timestamp = Instant.now.getEpochSecond.toString
```
AndyHWChung, line 67: this should be an empty string if it's not a further return

mikail-khan: fixed

mikail-khan:
@AndyHWChung can you re-review to make sure we are on the same page

AndyHWChung:
LGTM

PR #230, opened 2020-08-24
CPDP-2262: send email in chosen language

mikail-khan:


conf/application.conf
```diff
@@ -276,4 +276,6 @@ s3 {
     file-download-timeout = 2 minutes
     upstream-element-limit-scale-factor = 200 # do not adjust
     max-file-download-size-in-mb = 5
-}
\ No newline at end of file
+}
+
+httpHeadersWhitelist=["Accept-Language"]
```
AndyHWChung, line 281: we're supposed to use the term `allowList` now instead of `whitelist` and `denyList` instead of `backList`

mikail-khan: you can't - this is the key used in `http-verbs` and that can't change unless Plat Ops release a new version of the library

AndyHWChung: or is this even something we have control over?

AndyHWChung: would we not need `+=` if the config param comes from another libary?

mikail-khan: we don't have control over this. 
we don't need the += - this is a list specific to the service using the http-verbs library

AndyHWChung: :+1: didn't see your reply before writing my other comments

AndyHWChung:
LGTM

PR #237, opened 2020-09-08
CPDP-2430: delete section 3 of draft return upon amendment

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposals/service/returns/ReturnsService.scala
```diff
@@ -102,9 +104,77 @@ class DefaultReturnsService @Inject() (
...
+                supportingEvidenceAnswers = None
+              )
+            )
+        }
+
+    val cgtReference = submitReturnRequest.subscribedDetails.cgtReference
+    if (submitReturnRequest.originalReturnFormBundleId.isEmpty)
+      EitherT.pure(())
+    else {
+      import cats.instances.list._
```
mikail-khan, line 165: does this cause conflicts if places at the top of the file with the other imports?

AndyHWChung: yeah, in some other places it gets confused with the future cats instances in other for-comprehensions

test/uk/gov/hmrc/cgtpropertydisposals/models/Generators.scala
```diff
@@ -202,11 +202,27 @@ trait TaxEnrolmentGen { this: GenUtils =>
 
 }
 
-trait DraftReturnGen { this: GenUtils =>
+trait DraftReturnGen extends LowerPriorityDraftReturnGen { this: GenUtils =>
```
mikail-khan, line 205: why do you need to do this?

AndyHWChung: if you stick an implicit generator for `DraftReturn` with all the generators of all of the case classes which extend `DraftReturn` , the compiler will get confused when you try to summon `sample[DraftReturn]`. It will see the instance of `Gen[DraftReturn]` but it will also see the other generators which somehow it pieces together to something which also looks like `Gen[DraftReturn]`. Separating out seems to make the compiler happier so they don't conflict.

mikail-khan (approved):

PR #239, opened 2020-09-17
CPDP-2694: handle amended returns

mikail-khan:


app/uk/gov/hmrc/cgtpropertydisposals/repositories/returns/AmendReturnsRepository.scala
```diff
@@ -0,0 +1,105 @@
...
+      collectionName = "amend-returns",
+      mongo = component.mongoConnector.db,
+      domainFormat = SubmitReturnRequest.format
+    )
+    with AmendReturnsRepository
+    with CacheRepository[SubmitReturnRequest] {
+
+  //TODO: review parameters
+  val cacheTtl: FiniteDuration  = config.underlying.get[FiniteDuration]("mongodb.draft-returns.expiry-time").value
+  val maxDraftReturns: Int      = config.underlying.get[Int]("mongodb.draft-returns.max-draft-returns").value
```
mikail-khan, line 63: @AndyHWChung are these parameters applicable? How should the house-keeping be done for the entries in this collection?

AndyHWChung: with the way this is being used there shouldn't be a limit to the number of  returns returned for a cgt reference - to correctly display the "recently amended" content we have to know for certain whether the return was recently amended which means looking at everything in this collection for the cgt reference.

The `cacheTtle` needs to be separate from the one for draft returns. The draft return ttl is 30 days, but this ttl needs to be 24 hours. This ttl controls how long a user will see the "recently amended" version of the content on the homepage for after the return has been amended.

test/uk/gov/hmrc/cgtpropertydisposals/repositories/returns/AmendReturnsRepositorySpec.scala
```diff
@@ -0,0 +1,63 @@
...
+  val cgtReference = sample[CgtReference]
+
+  "DraftAmendReturnsRepository" when {
+    "inserting" should {
+      "create a new draft return successfully" in {
+        val submitReturnRequest = sample[SubmitReturnRequest]
+        await(repository.save(submitReturnRequest, cgtReference).value) shouldBe Right(())
+      }
+    }
+    "getting"   should {
```
AndyHWChung, line 50: could we store two sample `SubmitReturnRequest` with the same cgt referebce in this test to make sure we get both back?

app/uk/gov/hmrc/cgtpropertydisposals/repositories/returns/AmendReturnsRepository.scala
```diff
@@ -0,0 +1,104 @@
...
+  override def fetch(cgtReference: CgtReference): EitherT[Future, Error, List[SubmitReturnRequest]] =
+    EitherT(
+      preservingMdc {
+        get(cgtReference)
+      }
+    )
+
+  override def save(
+    submitReturnRequest: SubmitReturnRequest,
+    cgtReference: CgtReference
```
AndyHWChung, line 76: `cgtReference` isn't being used here

mikail-khan: nasty - the compiler should have flagged that (and Intellij!) and failed compilation

AndyHWChung:
LGTM

PR #242, opened 2020-09-23
CPDP-2800: Delta charges

rambabu-posa:


README.md
```diff
@@ -1,25 +1,3 @@
 # cgt-property-disposals
 
-For more context of Capital Gain Tax on UK Property (CGT), see [here](https://github.com/hmrc/cgt-property-disposals-frontend#cgt-property-disposals-frontend).
```
AndyHWChung, line 3: the README text should be restored

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/ReturnsService.scala
```diff
@@ -113,6 +123,79 @@ class DefaultReturnsService @Inject() (
     } yield returnResponse
   }
 
+  private def deltaCharge(
+    response: DesSubmitReturnResponse,
+    cgtReference: CgtReference,
+    fromDate: LocalDate,
+    toDate: LocalDate
+  )(implicit hc: HeaderCarrier): EitherT[Future, Error, Option[DeltaCharge]] =
+    response.ppdReturnResponseDetails.chargeReference match {
```
AndyHWChung, line 132: if the return is not an amend return we can return `None` straight away - delta charges only apply to amends

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/ReturnsService.scala
```diff
@@ -113,6 +123,79 @@ class DefaultReturnsService @Inject() (
...
+      case Some(chargeReference) =>
+        lazy val charge = for {
```
AndyHWChung, line 134: you don't need the `laazy val` here, you can just have the for-comprehension by itself

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/ReturnsService.scala
```diff
@@ -113,6 +123,79 @@ class DefaultReturnsService @Inject() (
...
+          desFinancialData <- getDesFinalcialData(cgtReference, fromDate, toDate)
+          charge           <- EitherT.fromEither[Future](findDeltaCharge(desFinancialData.financialTransactions, chargeReference))
+        } yield charge
+
+        charge
+      case _                     => EitherT.pure(None)
+    }
+
+  private def getDesFinalcialData(
```
AndyHWChung, line 143: this is the same code that appears elsewhere - it'd be better to make the original code not private and reuse it here

rambabu-posa: yes right, I had same plan to write a function and reuse at both places but forgot that one.

AndyHWChung:
is there a spec which tests that delta charge is returned properly? I can't see one but I could just be missing it

test/uk/gov/hmrc/cgtpropertydisposals/service/returns/ReturnsServiceSpec.scala
```diff
@@ -1460,6 +1460,281 @@ class ReturnsServiceSpec extends WordSpec with Matchers with MockFactory {
...
+
+        }
+
+        "there are two DesFinancialTransactions which matches return's charge reference" in {
+
+          val formBundleId    = "804123737752"
+          val chargeReference = "XCRG9448959757"
+
+          val submitReturnRequest    = sample[SubmitReturnRequest].copy(
+            originalReturnFormBundleId = None
```
AndyHWChung, line 1594: shouldn't this be `Some(..)`? delta charges should only apply when it's an amend return

AndyHWChung: how does this test pass if it's not an amend return and you're expecting a `DeltaCharge` at the end of it?

AndyHWChung:
LGTM

PR #312, opened 2022-04-20
[SM][DLS-5170] refactored ReturnService so that it uses the tax years…

SeanMurray1:
… from the config. Removed TaxYearExchanged from the backend as it is used in the frontend to show radio options in the view.


rambabu-posa (commented):
Im bit worried about removing _taxYearExchanged_
Please test it with QA

rambabu-posa (approved):
LGTM

PR #313, opened 2022-04-26
[SM][DLS-5173] replaced the feature flag with a date variable.

SeanMurray1:

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposals/service/returns/TaxYearService.scala
+++ b/app/uk/gov/hmrc/cgtpropertydisposals/service/returns/TaxYearService.scala
@@ -46,10 +46,10 @@ class TaxYearServiceImpl @Inject() (config: Configuration) extends TaxYearServic
 
   def taxYears: List[TaxYear] = {
     val taxYearsList = config.underlying.get[List[TaxYear]]("tax-years").value
-    if (LocalDate.now.isAfter(latestTaxYearLiveDate) || LocalDate.now.isEqual(latestTaxYearLiveDate))
-      taxYearsList
-    else
+    if (LocalDate.now.isBefore(latestTaxYearLiveDate))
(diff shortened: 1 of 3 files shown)


lutnos (changes requested):
The list of years need to be dynamic, Sio that it will changes the time progresses past the threshold

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/TaxYearService.scala
```diff
@@ -37,12 +37,16 @@ trait TaxYearService {
...
+  val taxYearLiveDateMonth = taxYearLiveDate.get[Int]("month").value
+  val taxYearLiveDateYear  = taxYearLiveDate.get[Int]("year").value
 
-  val taxYears: List[TaxYear] = {
+  val latestTaxYearLiveDate: LocalDate = LocalDate.of(taxYearLiveDateYear, taxYearLiveDateMonth, taxYearLiveDateDay)
+
+  def taxYears: List[TaxYear] = {
     val taxYearsList = config.underlying.get[List[TaxYear]]("tax-years").value
-    if (latestTaxYearEnabled)
+    if (LocalDate.now.isAfter(latestTaxYearLiveDate) || LocalDate.now.isEqual(latestTaxYearLiveDate))
```
lutnos, line 49: It would be more concise to invert this: use .isBefore and reverse the if ... else ...

lutnos (changes requested):
Just a couple of minor suggestions

app/uk/gov/hmrc/cgtpropertydisposals/service/returns/TaxYearService.scala
SeanMurray1: Yeah that makes scene

lutnos (approved):

PR #323, opened 2023-03-16
DLS-6922: Added new tax year for 2023-2024

johno1985:
- added new tax year and updated unit tests
 - removed unused OldTaxYearExchanged object
 - code formatting
 - use abstract Duration type and removed test library conversion
 - explicitly declared implicit formatters for Either type
 - removed old redundant TaxYearExhanged reads
 - updated README, sm -> sm2


lutnos (changes requested):
It is a bit difficult for anyone to review this, it is all very well having the copyright header change in a separate commit, but when there are many other commits there is no way in github to read through the diff without looking at all 291 files.  If there is just one commit other than the date then it would be straightforwards, look at it from the reviewers POV.  

Unless you can show me I am wrong, then I would suggest either doing the copyright date change as a separate PR and the rebasing on to that, or otherwise squashing the other commits - either would help.

johno1985:
> It is a bit difficult for anyone to review this, it is all very well having the copyright header change in a separate commit, but when there are many other commits there is no way in github to read through the diff without looking at all 291 files. If there is just one commit other than the date then it would be straightforwards, look at it from the reviewers POV.
> 
> Unless you can show me I am wrong, then I would suggest either doing the copyright date change as a separate PR and the rebasing on to that, or otherwise squashing the other commits - either would help.

I have extracted out the file header updates and import tidy up into separate PRs which have now been merged.

johno1985:
rebuild

PR #342, opened 2024-02-27
DLS-10086 | Add new tax year config entry

mahesh-hmrc:
Updated upper tax threshold and exemption amount accordingly for new tax year.

### DIFF
--- a/conf/tax_years.conf
+++ b/conf/tax_years.conf
@@ -16,13 +16,31 @@ latest-tax-year-go-live-date
     {
     day = 06
     month = 04
-    year = 2023
+    year = 2024
     }
 
 #########################################
 # IMPORTANT: Please add new tax year at top of this list.
 #########################################
 tax-years = [
+  {
+    start-year = 2024
+    annual-exempt-amount {
+      general              = 3000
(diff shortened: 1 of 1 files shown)


conf/tax_years.conf
```diff
@@ -16,13 +16,31 @@ latest-tax-year-go-live-date
     {
     day = 06
     month = 04
-    year = 2023
+    year = 2024
```
johno1985, line 19: I'm not sure if this should be changed right now, I think in the past we changed this separately and then released the change on April 6th

johno1985 (commented):
comment added

johno1985 (approved):

conf/tax_years.conf
mahesh-hmrc: hey! we should be safe to do so in a single change. Following in TaxYearService reflects what it's used for and is needed to make sure we return correct list of tax year configs. Thanks
https://github.com/hmrc/cgt-property-disposals/blob/main/app/uk/gov/hmrc/cgtpropertydisposals/service/returns/TaxYearService.scala#L54

PR #343, opened 2024-04-05
DLS-10054: Upgrade cgt-property-disposals to Play 3.0

johno1985:


test/uk/gov/hmrc/cgtpropertydisposals/models/Generators.scala
```diff
@@ -419,14 +407,12 @@ trait AddressGen extends AddressLowerPriorityGen {
   }
 
   implicit val countryGen: Gen[Country] = Gen.oneOf(Country.countryCodes.map(Country(_)))
-
 }
```
karolina-zaborska, line 410: I think those changes fit more properly under the cosmetics/refactoring rather than Play 3.0 upgrade but its just a suggestion for a future

conf/tax_years.conf
```diff
@@ -107,9 +107,9 @@ tax-years = [
     income-tax-higher-rate-threshold = 37500
     lettings-relief-max-threshold = 40000
     cgt-rates {
-      lower-band-residential      = 18
```
karolina-zaborska, line 110: I try to follow equal spacing (just like before changes) cause I believe it's clearer to read. Is any of the approaches the industry standard or is it just a matter of preferences? I'm just asking to know for my future PRs

karolina-zaborska (dismissed):
If you could address the BOT comment about uk.gov.hmrc.sbt-distributables that would be great. Other than that, I have nothing to complain about, on the contrary, your attention to detail is really something. Thanks!

conf/tax_years.conf
johno1985: I just used the auto formatter in Intellij

test/uk/gov/hmrc/cgtpropertydisposals/models/Generators.scala
johno1985: No probs, will try and keep them separate in the future

karolina-zaborska (approved):
LGTM

PR #349, opened 2024-08-19
DLS 10771

BetaDraconis:
- **DLS-10771: Add framework for connector testing with WireMock. Note, this includes a test for gateway timeouts which requires us to make the response take over 20 seconds to respond. In HTTP Client V2 we can customise the timeout time for this test to reduce it to a more reasonable number so the test doesn't take 20 seconds to run.**
- **DLS-10771: Move parsing from service to connector and refactor accordingly**


test/uk/gov/hmrc/cgtpropertydisposals/connectors/ConnectorSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+                          returnStatus: Int,
+                          expectedStatus: Int,
+                          returnBody: JsValue,
+                          expectedBody: JsValue,
+                          isErrorScenario: Boolean)
+
+  def testGET(downstreamUrl: String,
+              downstreamQueryParams: Map[String, String],
+              testScenarios: Seq[TestScenario],
+              connectorAction: () => EitherT[Future, Error, HttpResponse]): Unit = {
```
LAJW: I was considering building something like this, but ultimately decided against it. The abstraction layer hinders debugging and diagnosing test failures. And it's already pretty difficult to debug these things as is. I think a little bit of repetition in these tests is probably OK.

PR #370, opened 2025-07-01
DLS-11958 [CGTPD] Remove DMS Submission Poller

sandeep-k-soni:


mahesh-hmrc (approved):

mahesh-hmrc:
Hi @sandeep-k-soni , realised after the approval that with the poller now removed we no longer will need the `hmrc-mongo-work-item-repo` dependency. That transiently brings the mongo dependency, so, just replace that with standard `hmrc-mongo` dependency. Thanks

